PR #3, opened 2018-02-27
Update styles to latest version, move all css into same folder

gdbarnes:


gdbarnes:
My bad. Force of habit. 🙄

PR #6, opened 2018-02-28
On 172 updated login process

davidgouge:
Login process to match process flow required.
Includes change to Contact entity & appropriate migration.


davidgouge:
Thanks for the input. 

1) I only used an Orchestrator in this case as the Action was getting a little too busy. We shouldn't always use them as for very simple actions it's just another class to jump through that adds no value. 

2) Agree that handler should probably be renamed now. It has to do that update if the org already exists for orgs that were removed from register then were re added. The ability to change an orgs status through the api has been removed.

davidgouge:
@jcoxhead Thanks John. 

The decision is: be pragmatic about it.  Use whatever's clean and simple.  If you need to refactor out to an Orchestrator to clean up the Controller, then do that. If you don't, don't. :)  This doesn't need to be a 'from on high' decree about whether to use it all the time or not.  Just do what's clean and testable. :)

Posts fine for the time being as the update in it is a very edge-y case. And leave Put just in case.  It doesn't hurt it being there for now.

jcoxhead:
k thanks dave works for me we can always refactor later, but would like to minimise any rework  - I agree we don't want top down decisons - we just need to work together -  I am just concerned on consistency at the moment and perhaps overly concerned. Most importantly we just need agreement.  

Let's just make sure to keep a happy ship Alan is happy with this  and we'll move forward.

davidgouge:
Thanks @jcoxhead, can you approve then please?

PR #7, opened 2018-03-01
Remove Exposing Guids through view models

jcoxhead:
and using natural keys.

Plesae note: 

1). Some test are commented out ignore for now as I will be rewriting these to convert from mspec today.
All other test run ok.


jcoxhead:
Merge Remove Exposing Guids through view models into build_0 branch

jcoxhead:
merge Remove Exposing Guids through view models

PR #11, opened 2018-03-04
On 185 remove enums for statuses

jcoxhead:
Fixes to specflow tests, Remove enums for statuses and unit test conversion from mspec related Organisaton Tests


jcoxhead:
yep good spot thanks dave missed in the mellee - will do

PR #13, opened 2018-03-05
Updated GetAllOrganisations to not filter on Status

davidgouge:


jcoxhead:
Hi Dave - I am guessing you want to retrieve all records for organisation irregardless of whether they are deleted as the client will determine what to display. Should we do this also for all other queries like contacts?

Also what do we do for say get contact by username? if one is deleted and another one as been added as at the moment that is designed to returns only one record?

davidgouge:
Worry about the others when we get to them.

jcoxhead:
ok but Lets talk at the review - we should be able to work this out. 
We don't want to mix concepts. It will become a mess.

jcoxhead:
I will approve it for now.

PR #26, opened 2018-03-06
Merging master hotfixes into build_0.1

davidgouge:


davidgouge:
There's a conflict on the Api Startup.cs, but I can sort that once approved. :)

PR #76, opened 2018-03-15
On 254 make tests staleless

jcoxhead:
Dave I am assuming we are just working with master branch now. This pr relates to making the specflow tests stateless. The Organisation Tests will work, but at this point the Contact Tests are broken. Will work on those tommorow morning.


jcoxhead:
By the way it looks like we dont have rules to make sure one approver reviews on PR for master branch - for now will not merge and let you review.

PR #85, opened 2018-03-20
On 237 prototype for document conversion

jcoxhead:


jcoxhead:
Sure dave - agreed - I found one alternative from syncfusion the problem is that it's expensive $1,999 per developer seat. I'll keep my eye out if something comes up.

PR #259, opened 2018-05-23
On 553 may18 cert import changes

davidgouge:
Covers th following from the ticket:
1) To include the "Full Name" field on the Spreadsheet as "FullName" in the CertificateData json object, for all previously imported and new Certificates.
2) Where the ULN does not match to a current ILR record, and hence cannot lookup GivenNames and FamilyName, to decompose the Full Name as RIGHT Word (inclusive of Hyphens) = FamilyName, remainder = GivenNames.
3) Where the ULN is 9999999999 (which it should not be!) do not to attempt to match to ILR record 
4) Include all Certificates, including where Source = "Other" and not just Source ="ILR" - there may be an ILR record for these (as "Other" may / may not be correct); Step 2 will handle the completion of the name fields.
5) Also, note that Provider UKPRN is not always given for "Other", so should also add the "Provider Name" to CertificateData object as "ProviderName"
(I suggest we add "ProviderName" for all existing certificates, and then we can later get this via Provider Events API  using UKRPN for all newly created Certificates.
7) Handle "NOT APPLICABLE" in "Overall Grade" ==> "NO GRADE AWARDED" (Check exact wording)
8) The standard level data value has been imported as "LEVEL 2" and should be "2" - this needs to be corrected on all Certificates previously Imported.

### DIFF
--- a/LoadCertificateData.sql
+++ b/LoadCertificateData.sql
@@ -23,10 +23,6 @@
 --(1668 row(s) affected)
 --(1170 row(s) affected)
 
-
--- Update existing Certificate Records with FullName and a placeholder for ProviderName
-UPDATE Certificates SET CertificateData = JSON_MODIFY(JSON_MODIFY(CertificateData, '$.ProviderName', ''), '$.FullName', JSON_VALUE(CertificateData, '$.LearnerGivenNames') + ' ' + JSON_VALUE(CertificateData, '$.LearnerFamilyName'))
-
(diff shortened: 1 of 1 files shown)


davidgouge:
I going to assume by "StandardLevel" you mean "ProviderName" as that is the other new field in CertificateData.

davidgouge:
Also added in StandardLevel fix.

PR #300, opened 2018-06-11
On 566 add internal api endpoint to enable request for certificate to be reprinted

jcoxhead:


src/SFA.DAS.AssessorService.Data/CertificateRepository.cs
```diff
@@ -128,5 +127,28 @@ public CertificateRepository(AssessorDbContext context)
         {
             return await _context.CertificateLogs.Where(l => l.CertificateId == certificateId).OrderByDescending(l => l.EventTime).ToListAsync();
         }
+
```
davidgouge: This doesn't look like it sets the status of the Cert to Reprint.  In any case, you don't need a new method on the Repository like this.  Create a method to get a cert by those params.  Call that from handler, set new status on Cert and then pass it back to CertRepository.Update.  That will then take care of updating the Logs.

PR #304, opened 2018-06-12
On 569 print function to also handle certificates to be printed when in reprint status

jcoxhead:


src/SFA.DAS.AssessorService.Data/CertificateRepository.cs
```diff
@@ -69,21 +69,21 @@ private bool CheckCertificateData(Certificate certificate, string lastName, Date
...
-                return await _context.Certificates
-                    .Include(q => q.Organisation)
-                    .Where(q => q.Status == status)
-                    .ToListAsync();
-            }
+           {
+               return await _context.Certificates
+                   .Include(q => q.Organisation)
+                   .Where(x => statuses.Contains(x.Status))
+                   .ToListAsync();
```
davidgouge: We might have to revisit this as last time I did something like it, EF grabbed *all* of the certificates into memory and then did the Contains. 😭   Ok for now though. 🎉

jcoxhead: i'll profile the sql query it shouldn't  do that maybe something to do with ef core

davidgouge: Sure I read something recently that EF Core does these weird things.  But there was definitely part of the system where EF was only doing part of the Query on the SQL Server and the rest in memory.  Which was rubbish.

PR #322, opened 2018-06-18
On 616 print function stalling

jcoxhead:


davidgouge:
Reopened to merge current PROD into master.

PR #325, opened 2018-06-20
Dotnet 21 spike

davidgouge:
This was going to be a spike to have a quick look at perf, but it was so easy to do and the perf was so good, decided to keep it. 🎉 Also, the new Staff Web App will be 2.1 ('cause why not!? 🤷‍♂️ ) so it makes sense for all of our aspnet core apps to be 2.1.

Upgraded web and api to aspnet core 2.1 for perf improvements.

Would recommend a run through of the Cert process and then a check on Print flow to make sure all is still well.

### DIFF
--- a/src/SFA.DAS.AssessorService.Application.Api/SFA.DAS.AssessorService.Application.Api.csproj
+++ b/src/SFA.DAS.AssessorService.Application.Api/SFA.DAS.AssessorService.Application.Api.csproj
@@ -21,7 +21,7 @@
     <PackageReference Include="Microsoft.AspNetCore.App" />
     <!--<PackageReference Include="NBuilder" Version="5.0.0" />-->
     <PackageReference Include="Nlog.Web.AspNetCore" Version="4.5.4" />
-    <!--<PackageReference Include="SFA.DAS.NLog.Targets.Redis" Version="1.0.0.56651" />-->
+    <PackageReference Include="SFA.DAS.NLog.Targets.Redis" Version="1.1.5" />
     <PackageReference Include="StructureMap.Microsoft.DependencyInjection" Version="1.4.0" />
     <PackageReference Include="StructureMap" Version="4.6.1" />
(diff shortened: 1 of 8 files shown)


davidgouge:
OK, looks like our build servers aren't set up for 2.1 yet. 😆

davidgouge:
DevOps are looking into getting 2.1 onto build servers now that it's RTM. 👍

src/SFA.DAS.AssessorService.Application.Api/SFA.DAS.AssessorService.Application.Api.csproj
```diff
@@ -12,28 +12,26 @@
...
+    <PackageReference Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.3.0" />
+    <PackageReference Include="Microsoft.ApplicationInsights.NLogTarget" Version="2.6.4" />
+    <PackageReference Include="Microsoft.AspNetCore.App" />
+    <!--<PackageReference Include="NBuilder" Version="5.0.0" />-->
+    <PackageReference Include="Nlog.Web.AspNetCore" Version="4.5.4" />
+    <PackageReference Include="SFA.DAS.NLog.Targets.Redis" Version="1.1.5" />
     <PackageReference Include="StructureMap.Microsoft.DependencyInjection" Version="1.4.0" />
-    <PackageReference Include="Swashbuckle.AspNetCore" Version="1.1.0" />
+    <PackageReference Include="StructureMap" Version="4.6.1" />
+    <PackageReference Include="Swashbuckle.AspNetCore" Version="2.5.0" />
```
davidgouge: Was also a great opportunity to upgrade some nuget packages. 😁

davidgouge:
Ewan has installed core 2.1 on the build servers, so this could now do with a test!

davidgouge:
These upgrade changes will now be included in a different PR so this PR is no longer required.

PR #333, opened 2018-06-26
On 554 split address page

jcoxhead:


src/SFA.DAS.AssessorService.Api.Types/Models/CertificateAddressResponse.cs
```diff
@@ -0,0 +1,13 @@
+﻿using MediatR;
+
+namespace SFA.DAS.AssessorService.Api.Types.Models
+{
+    public class CertificateAddressResponse : IRequest
```
davidgouge: Does this need to implement IRequest? 🤔

src/SFA.DAS.AssessorService.Api.Types/Models/GetAddressesRequest.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Collections.Generic;
+
+namespace SFA.DAS.AssessorService.Api.Types.Models
+{
+    using MediatR;
```
davidgouge: Aww, look, a lost using statement. 😜

src/SFA.DAS.AssessorService.Api.Types/Models/CertificateAddressResponse.cs
davidgouge: In fact, do we need this class at all as it's nigh on identical to CertificateAddress? Maybe just return CertificateAddress all the way through and then there's no need for the Mapping.

src/SFA.DAS.AssessorService.Application/Handlers/Certificates/GetCertificateAddressesHandler.cs
```diff
@@ -0,0 +1,45 @@
...
+            _organisationQueryRepository = organisationQueryRepository;
+            _logger = logger;
+        }
+     
+        public async Task<List<CertificateAddressResponse>> Handle(GetAddressesRequest request, CancellationToken cancellationToken)
+        {
+            var contact = await _contactQueryRepository.GetContact(request.Username);
+            var organisation = await _organisationQueryRepository.Get(contact.EndPointAssessorOrganisationId);
+
+            var certificateAddresses = await _certificateRepository.GetCertificateAddresses(organisation.Id);
```
davidgouge: Could this just pass the username to GetCertificateAddresses and then do the link from Contact to Organisation in the sproc?  Save those two extra unneeded db calls? 😄

src/SFA.DAS.AssessorService.Data/CertificateRepository.cs
```diff
@@ -147,5 +153,13 @@ private bool CheckCertificateData(Certificate certificate, string lastName, Date
         {
             return await _context.CertificateLogs.Where(l => l.CertificateId == certificateId).OrderByDescending(l => l.EventTime).ToListAsync();
         }
+
+        public async Task<List<CertificateAddress>> GetCertificateAddresses(Guid organisationId)
+        {
+            var addresses = await _connection.QueryAsync<CertificateAddress>("GetRecentCertificateAddresses",
+                new {OrganisationId = organisationId},
+                commandType: CommandType.StoredProcedure);
```
davidgouge: Could you add the Stored Procedure to the Database project please? 🌟

src/SFA.DAS.AssessorService.Api.Types/Models/CertificateAddressResponse.cs
jcoxhead: Yep wasn't sure about passing the view model out of the repository. If your happy with this I have no issues -

davidgouge:
@jcoxhead Some extra comments have been hidden above because the file name has changed.  👍

src/SFA.DAS.AssessorService.Api.Types/Models/CertificateAddressResponse.cs
davidgouge: Yeah, there's nothing wrong with the Entity making it's way all the way out of the API. 👍

davidgouge:
I've removed the Requires Testing label as it's not ready yet with the UI bit missing.

src/SFA.DAS.AssessorService.Database/StoredProcedures/GetPreviousAddresses.sql
```diff
@@ -0,0 +1,30 @@
...
+  
+	SELECT DISTINCT TOP 10 
+	c.OrganisationId
+    ,JSON_VALUE([CertificateData],'$.ContactAddLine1') AS AddressLine1
+   ,JSON_VALUE([CertificateData],'$.ContactAddLine2') AS AddressLine2
+   ,JSON_VALUE([CertificateData],'$.ContactAddLine3') AS AddressLine3
+   ,JSON_VALUE([CertificateData],'$.ContactAddLine4') AS City
+      ,JSON_VALUE([CertificateData],'$.ContactPostCode') AS PostCode
+	  ,MAX(c.CreatedAt) As CreatedAt
+	FROM [SFA.DAS.AssessorService].[dbo].[Certificates] c
```
davidgouge: Could you remove the db qualifier from this line please? :)

src/SFA.DAS.AssessorService.Database/StoredProcedures/GetPreviousAddresses.sql
```diff
@@ -0,0 +1,30 @@
...
+	group by  c.OrganisationId, JSON_VALUE([CertificateData],'$.ContactAddLine1'),
+	JSON_VALUE([CertificateData],'$.ContactAddLine2'),
+	JSON_VALUE([CertificateData],'$.ContactAddLine3'),
+	JSON_VALUE([CertificateData],'$.ContactAddLine4'),
+	JSON_VALUE([CertificateData],'$.ContactPostCode'),
+	c.Status
+	having c.OrganisationId = @OrganisationId AND (c.Status = 'Submitted' OR c.Status = 'Printed' OR c.Status = 'Reprint')
```
davidgouge: Can you add an extra AND here please to filter out CreatedBy "Manual". 👍

src/SFA.DAS.AssessorService.Database/StoredProcedures/GetPreviousAddresses.sql
jcoxhead: yep missed that one ta

src/SFA.DAS.AssessorService.Application.Api/Controllers/CertificateQueryController.cs
```diff
@@ -40,6 +42,28 @@ public CertificateQueryController(IMediator mediator)
...
+        }
+
+        [HttpGet("addresses", Name = "GetPreviousAddresses")]
+        [SwaggerResponse((int)HttpStatusCode.OK, Type = typeof(List<CertificateAddress>))]
+        [SwaggerResponse((int)HttpStatusCode.InternalServerError, Type = typeof(ApiResponse))]
+        public async Task<IActionResult> GetPreviousAddresses([FromQuery] string userName)
+        {
+            var addresses = await _mediator.Send(new GetPreviousAddressesRequest { Username = userName });
+            if (addresses.Count == 0)
+                throw new ResourceNotFoundException();
```
davidgouge: @jcoxhead Can this just return an empty array or something please? Or at the very least handle the 404 on the web app end if there are no previous addresses (which is a perfectly valid state).

src/SFA.DAS.AssessorService.Application.Api/StartupConfiguration/Startup.cs
```diff
@@ -55,6 +55,12 @@ public IServiceProvider ConfigureServices(IServiceCollection services)
             {
                 //services.AddApplicationInsightsTelemetry();
 
+                services.AddMvc()
```
davidgouge: Adding all the Mvc() stuff into the pipeline seems a bit heavy handed to fix a json loop.  Anything else you can do to fix without dragging all that stuff in?

src/SFA.DAS.AssessorService.Application.Api/StartupConfiguration/Startup.cs
```diff
@@ -151,6 +157,7 @@ private IServiceProvider ConfigureIOC(IServiceCollection services)
                 option.UseSqlServer(Configuration.SqlConnectionString, options => options.EnableRetryOnFailure(3));
 
                 config.For<AssessorDbContext>().Use(c => new AssessorDbContext(option.Options));
+                config.For<IDbConnection>().Use(c => new SqlConnection(Configuration.SqlConnectionString));
```
davidgouge: Looks like this is being added twice.  Merge issue?

src/SFA.DAS.AssessorService.Api.Types/Models/Certificates/GeContacttPreviousAddressesRequest.cs
```diff
@@ -0,0 +1,10 @@
+﻿using MediatR;
+using SFA.DAS.AssessorService.Domain.Entities;
+
+namespace SFA.DAS.AssessorService.Api.Types.Models.Certificates
+{
+    public class GetContactPreviousAddressesRequest : IRequest<CertificateAddress>
```
davidgouge: Can you fix the typo on this file please?

jcoxhead: sure np

jcoxhead: done

src/SFA.DAS.AssessorService.Application.Api/StartupConfiguration/Startup.cs
jcoxhead: Yep will take a look - probably merge let's see, is easy enough to fix.

jcoxhead: done

src/SFA.DAS.AssessorService.Application.Api/StartupConfiguration/Startup.cs
jcoxhead: sorry dave I can't see one that can quickly be dealt with. If you can see another way by all means let me know. I don't think is heavy handed is just an argument to the serialiser to allow it to deal with recursive loops - the real problem cause by exposing entity and code first - we could change the definition to expose a view model ? this would mean some extra work

PR #335, opened 2018-06-27
Added Search Log

davidgouge:


davidgouge:
@reachash Please note that this is only for part 1 of the ticket as part 2 is handled by CertificateLogs.

PR #340, opened 2018-07-02
Data import codepage fixes

davidgouge:


davidgouge:
@alandburns I've updated the script to also use Source in the duplicate check.

PR #343, opened 2018-07-03
On 664 store provider name on certificate data

davidgouge:
Provider Name is now looked up and added to certificateData when Cert created


davidgouge:
@reachash It's not to do with different EPAOs.  It's the training Provider's name, not the EPAO's name that goes in that field.

PR #359, opened 2018-07-23
Removed file logging from nlog configs

davidgouge:


davidgouge:
This PR does not require testing.

PR #361, opened 2018-07-24
On 668 migrate assessment organisations data

MarkFCain:
This is to pull in a spreadsheet and load a database, rerunnable until source of truth of data switches from spreadsheet to database.  It is disposable code, so no tests added, and written to be functional.


src/SFA.DAS.AssessorService.Data/AssessmentOrgsImporter.cs
```diff
@@ -0,0 +1,185 @@
...
+            var spreadsheetDto = new AssessmentOrganisationsSpreadsheetDto();
+
+            var containerName = "assessmentorgs";
+            var initialiseContainer = new BlobContainerHelper(_configuration);
+
+            var container = initialiseContainer.GetContainer(containerName).Result;
+
+            var blob = container.GetBlockBlobReference(TemplateFile);
+
+            var memoryStream = new MemoryStream();
```
martin-hornsby: Put this inside a using statement

src/SFA.DAS.AssessorService.Data/AssessmentOrgsImporter.cs
```diff
@@ -0,0 +1,185 @@
...
+            }
+            return spreadsheetDto;
+
+        }
+
+        private void TearDownDatabase(StringBuilder progressStatus)
+        {
+            _logger.LogInformation($"Teardown process started; ");
+            var progress = _assessmentOrgsRepository.TearDownData();
+            progressStatus.Append((string)progress);
```
martin-hornsby: Bit picky here. I am assuming this is already a string then remove the redundant cast.

src/SFA.DAS.AssessorService.Data/AssessmentOrgsRepository.cs
```diff
@@ -0,0 +1,380 @@
...
+        private static string ConvertIntToSqlValueString(int? intToProcess)
+        {
+            return intToProcess == null
+                ? "null"
+                : $@"{intToProcess}";
+        }
+
+        private static string  ConvertDateToSqlValueString (DateTime? dateToProcess)
+        {           
+            if (dateToProcess == null)
```
martin-hornsby: Being picky again but I would use the .HasValue property

src/SFA.DAS.AssessorService.Data/AssessmentOrgsRepository.cs
```diff
@@ -0,0 +1,380 @@
...
+        private static string MakeStringSuitableForJson(string stringToProcess)
+        {
+            return stringToProcess == null
+                ? "null"
+                : $@"{stringToProcess.Replace("'", "''")}";
+        }
+
+        private static string ConvertIntToSqlValueString(int? intToProcess)
+        {
+            return intToProcess == null
```
martin-hornsby: Being picky again but I would use the .HasValue property

PR #375, opened 2018-08-06
On 784 get all organisations by standard

MarkFCain:
This can only be tech reviewed once ON-682, ON-764, ON-766, ON-765, ON-783 are in master


MarkFCain:
This will be merged as part of the ON-785 pull request

PR #380, opened 2018-08-09
On 758 update certificates batch

martin-hornsby:


martin-hornsby:
Consolidated branches

PR #381, opened 2018-08-09
On 760 submit certificates batch

martin-hornsby:


martin-hornsby:
Consolidated branches

PR #382, opened 2018-08-10
ON 797 Added level to History search

jcoxhead:


reachash:
Good to go into master

PR #384, opened 2018-08-10
On 759 delete certificates batch

martin-hornsby:


martin-hornsby:
Consolidated into branch ON-760_submit_certificates_batch

PR #388, opened 2018-08-13
Fix print function scheduling error

davidgouge:


davidgouge:
I've tested a deployment of this branch and the ScheduleRuns script operates as expected.

PR #395, opened 2018-08-16
On 828 staff ui schedule config

martin-hornsby:


src/SFA.DAS.AssessorService.Application.Api/Controllers/ScheduleController.cs
```diff
@@ -15,10 +15,10 @@ public ScheduleController(IScheduleRepository scheduleRepository)
             _scheduleRepository = scheduleRepository;
         }
 
-        [HttpGet("api/v1/schedule", Name="GetNextScheduleToRunNow")]
-        public async Task<IActionResult> GetNextScheduleToRunNow(int scheduleType)
+        [HttpGet("api/v1/schedule", Name = "GetScheduleRun")]
+        public async Task<IActionResult> GetScheduleRun(Guid scheduleRunId)
         {
-            var scheduleRun = await _scheduleRepository.GetNextScheduleToRunNow(scheduleType);
+            var scheduleRun = await _scheduleRepository.GetScheduleRun(scheduleRunId);
```
davidgouge: Looks like you'rve removed "GetNextScheduleToRunNow"? I know it's clumsily named but it doesn't deserve to die. ;)

src/SFA.DAS.AssessorService.Application/Interfaces/IScheduleRepository.cs
```diff
@@ -1,11 +1,14 @@
...
+using System.Collections.Generic;
 using System.Threading.Tasks;
 using SFA.DAS.AssessorService.Domain.Entities;
 
 namespace SFA.DAS.AssessorService.Application.Interfaces
 {
     public interface IScheduleRepository
     {
+        Task<ScheduleRun> GetScheduleRun(Guid scheduleRunId);
+        Task<IEnumerable<ScheduleRun>> GetAllScheduleRun(int scheduleType);
```
davidgouge: Hmmm, and GetNextScheduleToRunNow is missing from the Repo.  Did you pull latest master before starting? :)

src/SFA.DAS.AssessorService.Web.Staff/Controllers/ScheduleConfigController.cs
```diff
@@ -0,0 +1,165 @@
...
+            return RedirectToAction("Index");
+        }
+    }
+
+    public enum ScheduleInterval
+    {
+        Hourly = 60,
+        Daily = 1440,
+        Weekly = 10080
+    }
```
davidgouge: Pop this out into Domain.

src/SFA.DAS.AssessorService.Web.Staff/Controllers/ScheduleConfigController.cs
```diff
@@ -0,0 +1,165 @@
...
+
+    public enum ScheduleJobType
+    {
+        [Display(Name="Print Run")]
+        PrintRun = 1
+    }
```
davidgouge: Same with this.

src/SFA.DAS.AssessorService.Web.Staff/Controllers/ScheduleConfigController.cs
```diff
@@ -0,0 +1,165 @@
...
+
+    public class ScheduleConfigViewModel
+    {
+        public Guid Id { get; set; }
+        public DateTime RunTime { get; set; }
+        public string Interval { get; set; }
+        public ScheduleInterval? ScheduleInterval { get; set; }
+        public bool IsRecurring { get; set; }
+        public ScheduleJobType ScheduleType { get; set; }
+    }
```
davidgouge: Put this into a ViewModels folder in its own class. 👍

src/SFA.DAS.AssessorService.Application.Api/Controllers/ScheduleController.cs
martin-hornsby: Ah yes, this is because this function now grabs the details of the specified `scheduleRunId` rather than its type. The reason behind that is to give the caller more control over what they would like to view/modify

src/SFA.DAS.AssessorService.Web.Staff/Startup.cs
```diff
@@ -63,6 +64,9 @@ public IServiceProvider ConfigureServices(IServiceCollection services)
             });
 
             services.AddMvc().SetCompatibilityVersion(CompatibilityVersion.Version_2_1);
+            services.AddMvc()
+                .AddMvcOptions(m => m.ModelMetadataDetailsProviders.Add(new HumanizerMetadataProvider()))
+                .SetCompatibilityVersion(CompatibilityVersion.Version_2_1);
```
davidgouge: Looks like .AddMvc() is being called twice now.  Not sure if that makes any difference, but still.  🤷‍♂️

src/SFA.DAS.AssessorService.Web.Staff/Infrastructure/ApiClient.cs
```diff
@@ -26,25 +26,48 @@ public ApiClient(HttpClient client, ILogger<ApiClient> logger, ITokenService tok
             _tokenService = tokenService;
         }
 
-        private async Task<T> Get<T>(string uri)
+        protected async Task<T> Get<T>(string uri)
```
davidgouge: Any reason why this was changed to `protected`? 😁

src/SFA.DAS.AssessorService.Web.Staff/Startup.cs
davidgouge: Hoisted by my own petard!

PR #398, opened 2018-08-17
On 821 stui view audit history

gdbarnes:


src/SFA.DAS.AssessorService.Web.Staff/Views/Search/Select.cshtml
```diff
@@ -21,103 +21,115 @@
...
+                        <td>@Model.Learner.Standard (@Model.Learner.StandardCode)</td>
+                    </tr>
+                    <tr>
+                        <td><span class="bold">Option</span></td>
+                        <td>@Model.Learner.Option</td>
+                    </tr>
+                    <tr>
+                        <td><span class="bold">Level</span></td>
+                        <td>@Model.Learner.Level</td>
+                    </tr>
```
davidgouge: @gdbarnes It looks like you've changed the order of Standard, Level, Option to Standard, Option, Level.  Is this intentional? :)

gdbarnes: I didn't think I had. Probably something to do with fixing the merge conflicts.

PR #401, opened 2018-08-27
On 811 handle comms problems with external ap is better

jcoxhead:


davidgouge:
image
Tested successfully with a 50% 500 autoresponder on the external API.  Failed on the first call and then automatically retried for the second. 👍

reachash:
Tested it and all looks good.

PR #411, opened 2018-09-03
On 867 search for organisations

MarkFCain:


src/SFA.DAS.AssessorService.Data/RegisterQueryRepository.cs
```diff
@@ -255,5 +255,43 @@ public async Task<IEnumerable<OrganisationStandardPeriod>> GetOrganisatonStandar
                 return await connection.QueryAsync<OrganisationStandardPeriod>(sql);
             }
         }
+
+        public async Task<IEnumerable<AssessmentOrganisationSummary>> GetAssessmentOrganisationsByUkprn(string ukprn)
+        {
+            var connectionString = _configuration.SqlConnectionString;
+            if (!int.TryParse(ukprn, out int ukrpnNumeric))
```
davidgouge: ukrpnNumeric should be ukprnNumeric. 😸

src/SFA.DAS.AssessorService.Data/RegisterQueryRepository.cs
```diff
@@ -255,5 +255,43 @@ public async Task<IEnumerable<OrganisationStandardPeriod>> GetOrganisatonStandar
...
+            {
+                return new List<AssessmentOrganisationSummary>();
+            }
+            using (var connection = new SqlConnection(connectionString))
+            {
+                if (connection.State != ConnectionState.Open)
+                    await connection.OpenAsync();
+                var assessmentOrganisationSummaries = await connection.QueryAsync<AssessmentOrganisationSummary>($@"select EndPointAssessorOrganisationId as Id, EndPointAssessorName as Name, EndPointAssessorUkprn as ukprn from [Organisations] where EndPointAssessorUkprn = {ukrpnNumeric}");
```
davidgouge: Should these sql queries be parameterized and pass an obj instead of using string interpolation? 🍸

src/SFA.DAS.AssessorService.Api.Types/Models/SearchAssessmentOrganisationsRequest.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Collections.Generic;
+using MediatR;
+using SFA.DAS.AssessorService.Api.Types.Models.AO;
+
+namespace SFA.DAS.AssessorService.Api.Types.Models
+{
+    public class SearchAssessmentOrganisationsRequest : IRequest<List<AssessmentOrganisationSummary>>
+    {
+        public string Searchstring { get; set; }
```
jcoxhead: Just wondering about the naming of this ? almost like hungarian notation - I would remove string on SearchString.

src/SFA.DAS.AssessorService.Application.Api.UnitTests/Controllers/Register/Query/SearchOrganisationBySearchStringTests.cs
```diff
@@ -0,0 +1,82 @@
...
+        public void ResultsAreOfTypeListAssessmentOrganisationDetails()
+        {
+            ((OkObjectResult)_result).Value.Should().BeOfType<List<AssessmentOrganisationSummary>>();
+        }
+        [Test]
+        public void ResultsMatchExpectedListOfAssessmentOrganisationDetails()
+        {
+            var organisations = ((OkObjectResult)_result).Value as List<AssessmentOrganisationSummary>;
+            organisations.Count.Should().Be(2);
+            organisations.Should().Contain(_assOrgDetails1);
```
jcoxhead: Picky point but I am wondering whether or not we should have a single test per assert?

src/SFA.DAS.AssessorService.Application.Api.UnitTests/Validators/Register/EpaORganisationSearchValidatorTests.cs
```diff
@@ -0,0 +1,47 @@
...
+        [TestCase("EPA000", false, "checking there is at least 4 numbers after the 'epa'")]
+        [TestCase("epa0000", true, null)]
+        [TestCase("EPA0000A", false, null)]
+        [TestCase("EPA000000000", true, null)]
+        [TestCase("EPA0000000000", false, "checking it is not greated that 12 characters in all")]
+        [TestCase("stuff", false, null)]
+        [TestCase("", false, null)]
+        public void CheckEpaOrganisationIdIsAcceptable(string organisationIdToCheck, bool matching, string message)
+        {
+            var res = _validator.IsValidEpaOrganisationId(organisationIdToCheck);
```
jcoxhead: Not sure about the use of the word res ? if it;s a result maybe should be called result

src/SFA.DAS.AssessorService.Application.UnitTests/Handlers/Register/Query/RegisterQuerySearchAssessmentOrganisationsTests.cs
```diff
@@ -0,0 +1,124 @@
...
+            organisations.Count.Should().Be(2);
+            organisations.Should().Contain(_assessmentOrganisationDetails1);
+            organisations.Should().Contain(_assessmentOrganisationDetails2);
+        }
+
+        [Test]
+        public void SearchAssessmentOrganisationsWithValidukprn()
+        {
+            const string searchstring = "12345678";
+            var request = new SearchAssessmentOrganisationsRequest { Searchstring = searchstring };
```
jcoxhead: General comment but would seperate out asserts via AAA into seperate methods like many of the other tests, 
and maybe mimic GivenWhen syntax. Maybe we should discuss with everyone whether or not we need consistency with tests. I'm not sure for example whether or not we should have a template for tests or whether or not to consider Xunit for example?

src/SFA.DAS.AssessorService.Application/Interfaces/IEpaOrganisationSearchValidator.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace SFA.DAS.AssessorService.Application.Interfaces
+{
+    public interface IEpaOrganisationSearchValidator
+    {
+        bool IsValidEpaOrganisationId(string organisationIdToCheck);
+
```
jcoxhead: Not sure vertical space between methods adds clarity

src/SFA.DAS.AssessorService.Data.IntegrationTests/GetAssessmentOrganisationsByNameTests.cs
```diff
@@ -0,0 +1,83 @@
...
+
+namespace SFA.DAS.AssessorService.Data.IntegrationTests
+{
+    public class GetAssessmentOrganisationsByNameTests : TestBase
+    {
+        private OrganisationModel _organisation1;
+        private OrganisationModel _organisation2;
+        private readonly DatabaseService _databaseService = new DatabaseService();
+        private RegisterQueryRepository _repository;
+        private string _orgId1;
```
jcoxhead: would fully qualify name _organisationId

src/SFA.DAS.AssessorService.Application.Api.UnitTests/Controllers/Register/Query/SearchOrganisationBySearchStringTests.cs
davidgouge: 🤔 Personally I think in this case it's valid as the two asserts are v closely related.  But generally, I'd agree. 👍

src/SFA.DAS.AssessorService.Data/RegisterQueryRepository.cs
MarkFCain: yep, typo

src/SFA.DAS.AssessorService.Application.Api.UnitTests/Controllers/Register/Query/SearchOrganisationBySearchStringTests.cs
jcoxhead: Yep 👍

src/SFA.DAS.AssessorService.Data/RegisterQueryRepository.cs
MarkFCain: yes agree, they should be paramaterized without fail for dapper.  Extra especially when they are strings.

src/SFA.DAS.AssessorService.Api.Types/Models/SearchAssessmentOrganisationsRequest.cs
MarkFCain: I just meant it as a usage of the expression 'search string', and I had precedence from other parts of the codebase using it the same way (not that this is a good reason?).  I personally think we do need to guard against adding type to variables, but 'string' here isn't there for indicating type, but meaning.

src/SFA.DAS.AssessorService.Application.Api.UnitTests/Controllers/Register/Query/SearchOrganisationBySearchStringTests.cs
MarkFCain: Hmm, this is tricky, because I'm really only concerned that the list is full and contains all elements.  As ordering is not part of the criteria, it should pass whatever the order of the items.  If ordering WERE part of the criteria, I'd do an assert comparing the lists, but as it's not I can't do that.... so I see the 3 asserts here are actually checking one thing, and by breaking them apart into separate tests, I'm fracturing what I'm trying to test.... does that make sense?  If any assert fails, it fails my test rather than my 'tests', as it were.

src/SFA.DAS.AssessorService.Application.Api.UnitTests/Validators/Register/EpaORganisationSearchValidatorTests.cs
MarkFCain: I was being lazy, will update.

MarkFCain: PS. I've done this all over the place, so there'll be some new lines of changed code dragged in 🗡

MarkFCain: Well, I'm a strong believer in the Boy Scout Rule... :p

src/SFA.DAS.AssessorService.Application/Interfaces/IEpaOrganisationSearchValidator.cs
MarkFCain: No it doesnt :D

src/SFA.DAS.AssessorService.Data.IntegrationTests/GetAssessmentOrganisationsByNameTests.cs
MarkFCain: again, me being lazy. Will fix where I've done it.

src/SFA.DAS.AssessorService.Application.UnitTests/Handlers/Register/Query/RegisterQuerySearchAssessmentOrganisationsTests.cs
MarkFCain: big discussion to be had another time :)

jcoxhead: 👍

PR #413, opened 2018-09-04
On 697 staff ui azure keygen

martin-hornsby:


src/SFA.DAS.AssessorService.Web/Controllers/ExternalApiController.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            var result = await _apiClient.CreateUser(ukprn, username);
+            return RedirectToAction("Index");
+        }
+
+        [HttpPost]
+        public async Task<IActionResult> RemoveAccess(AzureUser viewModel)
+        {
+            var ukprn = _contextAccessor.HttpContext.User.Claims.First(c => c.Type == "http://schemas.portal.com/ukprn")?.Value;
+            var username = _contextAccessor.HttpContext.User.Claims.First(c => c.Type == "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/upn")?.Value;
```
davidgouge: See these a lot throughout .Web.  I wonder if there's a middleware we can add to auto inject this stuff into controllers?  Similar to the ukprn / email thing in the External API? 🤔

src/SFA.DAS.AssessorService.Application.Api.Client/Azure/AzureApiClient.cs
```diff
@@ -0,0 +1,243 @@
...
+                user = await PostPutRequestWithResponse<CreateAzureUserRequest, AzureUser>(httpRequest, request);
+            }
+
+            if (user.AzureId != null)
+            {
+                var subscription = await SubscribeUserToProduct(user.Id, _productId);
+                user.Subscriptions.Add(subscription);
+            }
+
+            //await _organisationsApiClient.Update(new UpdateOrganisationRequest
```
jcoxhead: Picky point:- Commented out code for me should be assigned to the source control system
might be worthwhile in group discussion.

src/SFA.DAS.AssessorService.Application.Api.Client/Azure/AzureApiClient.cs
```diff
@@ -0,0 +1,243 @@
...
+
+            if (user != null)
+            {
+                using (var httpRequest = new HttpRequestMessage(HttpMethod.Delete, $"/users/{userId}?api-version=2017-03-01&deleteSubscriptions=true"))
+                {
+                    httpRequest.Headers.Add("If-Match", "*");
+                    await Delete(httpRequest);
+                }
+
+                //var organisation = await _organisationsApiClient.Get(user.Note);
```
jcoxhead: Same as previous  comment ...

src/SFA.DAS.AssessorService.Application.Api.Client/Azure/AzureApiClient.cs
```diff
@@ -0,0 +1,243 @@
...
+                //    EndPointAssessorOrganisationId = organisation.EndPointAssessorOrganisationId,
+                //    EndPointAssessorUkprn = organisation.EndPointAssessorUkprn,
+                //    ApiEnabled = false,
+                //    ApiUser = null
+                //});
+            }
+
+        }
+
+        public async Task<IEnumerable<AzureProduct>> ListProducts()
```
jcoxhead: Not sure if we need localisation with resource strings? for this project.

src/SFA.DAS.AssessorService.Application.Api.Client/Azure/AzureApiClientBase.cs
```diff
@@ -0,0 +1,159 @@
...
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
+using SFA.DAS.AssessorService.Application.Api.Client.Clients;
+using SFA.DAS.AssessorService.Application.Api.Client.Exceptions;
+
+namespace SFA.DAS.AssessorService.Application.Api.Client.Azure
+{
+    public abstract class AzureApiClientBase : IDisposable
```
jcoxhead: We seem to have the same class used in a few projects now - maybe we should rationalise this?

src/SFA.DAS.AssessorService.Web.Staff/Controllers/ExternalApiController.cs
```diff
@@ -0,0 +1,73 @@
...
+using System.Threading.Tasks;
+
+namespace SFA.DAS.AssessorService.Web.Staff.Controllers
+{
+    [Authorize]
+    public class ExternalApiController : Controller
+    {
+        private readonly ILogger<ExternalApiController> _logger;
+        private readonly IAzureApiClient _apiClient;
+        private readonly ISessionService _sessionService;
```
jcoxhead: _sessionService not used?

src/SFA.DAS.AssessorService.Web/Controllers/ExternalApiController.cs
```diff
@@ -0,0 +1,77 @@
...
+
+namespace SFA.DAS.AssessorService.Web.Controllers
+{
+    [Authorize]
+    public class ExternalApiController : Controller
+    {
+        private readonly ILogger<ExternalApiController> _logger;
+        private readonly IHttpContextAccessor _contextAccessor;
+        private readonly IAzureApiClient _apiClient;
+        private readonly ISessionService _sessionService;
```
jcoxhead: Unused private variables

src/SFA.DAS.AssessorService.Web/Controllers/ExternalApiController.cs
```diff
@@ -0,0 +1,77 @@
...
+            return RedirectToAction("Index");
+        }
+
+        [HttpPost]
+        public async Task<IActionResult> RemoveAccess(AzureUser viewModel)
+        {
+            var ukprn = _contextAccessor.HttpContext.User.Claims.First(c => c.Type == "http://schemas.portal.com/ukprn")?.Value;
+            var username = _contextAccessor.HttpContext.User.Claims.First(c => c.Type == "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/upn")?.Value;
+
+            if (!string.Equals(viewModel.State, "blocked"))
```
jcoxhead: Use of magic strings maybe we should use const

martin-hornsby: Feels a bit superfluous in this case. It is an isolated, the value won't ever change and it's self explanatory as to what it's being used for.

src/SFA.DAS.AssessorService.Application.Api.Client/Azure/AzureApiClient.cs
martin-hornsby: Localisation hasn't been spec'ed as of yet. @davidgouge could you let me know if you want this.

src/SFA.DAS.AssessorService.Web/Controllers/ExternalApiController.cs
martin-hornsby: I tried to add this but for some reason the middleware couldn't load up the HttpContext correctly

PR #414, opened 2018-09-05
On 804 api put existing epao organisation

MarkFCain:


src/SFA.DAS.AssessorService.Data/RegisterRepository.cs
```diff
@@ -37,13 +37,31 @@ public async Task<string> CreateEpaOrganisation(EpaOrganisation org)
                     "INSERT INTO [Organisations] ([Id],[CreatedAt],[EndPointAssessorName],[EndPointAssessorOrganisationId], " +
                     "[EndPointAssessorUkprn],[Status],[OrganisationTypeId],[OrganisationData]) " +
                     $@"VALUES (@id, getdate(), @name, @organisationId, @ukprn, 'New', @organisationTypeId,  @orgData)",
```
davidgouge: GetUtcDate() ? 😀

src/SFA.DAS.AssessorService.Data/RegisterRepository.cs
```diff
@@ -37,13 +37,31 @@ public async Task<string> CreateEpaOrganisation(EpaOrganisation org)
...
+        {
+            using (var connection = new SqlConnection(_configuration.SqlConnectionString))
+            {
+                if (connection.State != ConnectionState.Open)
+                    await connection.OpenAsync();
+
+                var orgData = JsonConvert.SerializeObject(org.OrganisationData);
+
+                connection.Execute(
+                    "UPDATE [Organisations] SET [UpdatedAt] = GetDate(), [EndPointAssessorName] = @Name, " +
```
davidgouge: GetUtcDate() here too? :)

src/SFA.DAS.AssessorService.Application.Api.UnitTests/Controllers/Register/Command/UpdateOrganisationNotFoundTests.cs
```diff
@@ -0,0 +1,101 @@
...
+namespace SFA.DAS.AssessorService.Application.Api.UnitTests.Controllers.Register.Command
+{
+    [TestFixture]
+    public class UpdateOrganisationNotFoundTests
+    {
+        private static RegisterController _controller;
+        private static Mock<IMediator> _mediator;
+        private static Mock<ILogger<RegisterController>> _logger;
+        private object _result;
+        private EpaOrganisation _returnedOrganisation;
```
martin-hornsby: _returnedOrganisation isn't being used. Are you missing a test?

src/SFA.DAS.AssessorService.Data/RegisterRepository.cs
MarkFCain: oh yes

src/SFA.DAS.AssessorService.Data/RegisterRepository.cs
MarkFCain: yes again

src/SFA.DAS.AssessorService.Application.Api.UnitTests/Controllers/Register/Command/UpdateOrganisationNotFoundTests.cs
MarkFCain: I is a neglected leftover when the original spec had the organisation being returned with the update, rather than an OK and an organisation ID. I've removed this now.  Eagle Eyes!

PR #421, opened 2018-09-13
ON-668b changes so far

MarkFCain:
I wanted the code change to be distinct so this is comparing the changes I've made with the original import code - basically, I had to add and populate a new field (ContactId) to OrganisationStandard table.  If you guys approve it, I'll actually merge this back into ON-668 and then it can go forward with testing. Ta.


MarkFCain:
This is being remerged back into ON-668 locally, PR no longer needed.

PR #422, opened 2018-09-14
On 867b dealing with special characters

MarkFCain:
This is reworking to handle the crazy characters that might come in.  Once you approve, I will push this back into ON-668 and tell the testers to redo


davidgouge:
Am I being dense, but....  why is there _so much_ what looks like master stuff being merged from 867b into 867?  I was expecting just the special characters changes.

MarkFCain:
I merged both with master and pushed and that added all the extra stuff - I've no idea why, because changes should be same in the ON-867 and ON-876b branch.... however, I've now merged ON-867 to ON-867b, so you should only see the 5 files changed by this... phew...

PR #424, opened 2018-09-18
On 805 api post new organisation standard

MarkFCain:


src/SFA.DAS.AssessorService.Application.Api/Controllers/RegisterController.cs
```diff
@@ -93,5 +94,41 @@ public async Task<IActionResult> UpdateEpaOrganisation([FromBody] UpdateEpaOrgan
...
+        {
+            try
+            {
+                _logger.LogInformation("Creating new Organisation Standard");
+                var result = await _mediator.Send(request);
+                return Ok(new EpaOrganisationStandardResponse(result));
+            }
+            catch (NotFound ex)
+            {
+                _logger.LogError($@"Record is not available for organisation / standard: [{request.OrganisationId}, {request.StandardCode}]");
```
jcoxhead: There is some middleware we have added - See ErrorWrappingMiddleware which probably could do with extending to try to remove noise in the controllers cause by using try / catch blocks. Might be worthwhile taking a look.

src/SFA.DAS.AssessorService.Application.Api/Validators/EpaOrganisationValidator.cs
```diff
@@ -3,29 +3,38 @@
...
-        public string ErrorMessageOrganisationIdTooLong { get; } = "The length of the organisation Id is too long; ";
-        public string ErrorMessageOrganisationNameEmpty { get; } = "There is no organisation name; ";
-        public string ErrorMessageOrganisationIdAlreadyUsed { get; } = "There is already an entry for this organisation Id; ";
-        public string ErrorMessageUkprnAlreadyUsed { get; } = "There is already an organisation with this ukprn; ";
-        public string ErrorMessageOrganisationTypeIsInvalid { get; } = "There is no organisation type with this Id; ";
-        public string ErrorMessageAnotherOrganisationUsingTheUkprn { get; } = "The ukprn entered is already used by another organisation; ";
-        public string ErrorMessageUkprnIsInvalid { get; } = "The ukprn is not the correct format or length; ";
-        public string ErrorMessageOrganisationNotFound { get; } = "There is no organisation for the this organisation Id; ";
+        private string ErrorMessageNoOrganisationId { get; } = "There is no organisation Id; ";
+        private string ErrorMessageOrganisationIdTooLong { get; } = "The length of the organisation Id is too long; ";
```
jcoxhead: Not sure if we need to use localisation for string - in the web app we have doe this and also in the Api Controller to date.

src/SFA.DAS.AssessorService.Application.Api.UnitTests/SFA.DAS.AssessorService.Application.Api.UnitTests.csproj
```diff
@@ -19,4 +19,7 @@
     <ProjectReference Include="..\SFA.DAS.AssessorService.Application.Api\SFA.DAS.AssessorService.Application.Api.csproj" />
     <ProjectReference Include="..\SFA.DAS.AssessorService.Data\SFA.DAS.AssessorService.Data.csproj" />
   </ItemGroup>
+  <ItemGroup>
+    <None Remove="Controllers\Register\Command\CreateNewOrganisationTests.cs.orig" />
```
martin-hornsby: See if it is possible to remove this entry from the csproj to keep it tidy :)

MarkFCain:
Part of testing on ON-938 so will be merged in with that, so PR no longer required.

PR #427, opened 2018-09-20
On 692 privately funded apprentice certificate requests digital solution

gdbarnes:


src/SFA.DAS.AssessorService.Application/Handlers/Private/StartPrivateCertificateHandler.cs
```diff
@@ -0,0 +1,105 @@
...
+            //    provider = await _assessmentOrgsApiClient.GetProvider(ilr.UkPrn);
+            //}
+            //catch (Exception)
+            //{
+            //    // see whether there are any previous certificates with this ukrpn and a ProviderName....
+            //    var previousProviderName = await _certificateRepository.GetPreviousProviderName(ilr.UkPrn);
+            //    provider = previousProviderName != null
+            //        ? new Provider { ProviderName = previousProviderName }
+            //        : new Provider { ProviderName = "Unknown" };
+            //}
```
davidgouge: These commented lines can be deleted now. 😁 
And that `"CreateNewCertificate Before Get Standard from API"` log entry.

jcoxhead: yep:-  I think we are looking at an outdated copy - the comments where removed on friday and I can also see in my local copy they are removed.

jcoxhead: Hopefully should be ok now.

jcoxhead: amend bits done

PR #428, opened 2018-09-24
On 806 api put organisation standard

MarkFCain:
Hi chaps, As this is still waiting for ON-805 to be tested, I decided to push this into TR but comparing it to 805.  (I want to get this testing done as it might be a blocker, so I'll be asking testers to test ON-805, ON-806 and ON-938 together)


MarkFCain:
Part of testing on ON-938 so will be merged in with that, so PR no longer required.

PR #429, opened 2018-09-24
On 938 cleanse data on register end point parameters

MarkFCain:
I've requested a TR against ON-806 as I want these tested together, but didn't want to give you guys two lots of code mixed up to review.

I'm particularly keen if anyone has comments about any extra special characters to add in the 'SpecialCharacterCleanserService.cs'

Ta

### DIFF
--- a/src/SFA.DAS.AssessorService.Application/Handlers/EpaOrganisationHandlers/CreateEpaOrganisationHandler.cs
+++ b/src/SFA.DAS.AssessorService.Application/Handlers/EpaOrganisationHandlers/CreateEpaOrganisationHandler.cs
@@ -19,11 +19,13 @@ namespace SFA.DAS.AssessorService.Application.Handlers.EpaOrganisationHandlers
         private readonly ILogger<CreateEpaOrganisationHandler> _logger;
         private readonly IEpaOrganisationValidator _validator;
         private readonly IEpaOrganisationIdGenerator _organisationIdGenerator;
-
-        public CreateEpaOrganisationHandler(IRegisterRepository registerRepository, IEpaOrganisationValidator validator, IEpaOrganisationIdGenerator orgIdGenerator, ILogger<CreateEpaOrganisationHandler> logger)
+        private readonly ISpecialCharacterCleanserService _cleanser;
+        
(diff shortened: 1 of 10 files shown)


src/SFA.DAS.AssessorService.Application/Handlers/EpaOrganisationHandlers/CreateEpaOrganisationHandler.cs
```diff
@@ -53,6 +55,18 @@ public async Task<string> Handle(CreateEpaOrganisationRequest request, Cancellat
...
+            request.Name = _cleanser.CleanseStringForSpecialCharacters(request.Name?.Trim());           
+            request.LegalName = _cleanser.CleanseStringForSpecialCharacters(request.LegalName?.Trim());
+            request.WebsiteLink = _cleanser.CleanseStringForSpecialCharacters(request.WebsiteLink?.Trim());
+            request.Address1 = _cleanser.CleanseStringForSpecialCharacters(request.Address1?.Trim());
+            request.Address2 = _cleanser.CleanseStringForSpecialCharacters(request.Address2?.Trim());
+            request.Address3 = _cleanser.CleanseStringForSpecialCharacters(request.Address3?.Trim());
+            request.Address4 = _cleanser.CleanseStringForSpecialCharacters(request.Address4?.Trim());
+            request.Postcode = _cleanser.CleanseStringForSpecialCharacters(request.Postcode?.Trim());
+        }
+
```
davidgouge: Could this be a fancy bit of reflection to process each property on CreateEpaOrganisationRequest? Would save all these similar lines and would also catch any potential new fields in the future?

MarkFCain: I'm usually suspicious of reflection on the whole cos it can be cryptic about what's going on.  I did consider it briefly but I've seen reflection mess up understanding time and again and therefore be a drag on refactoring and reworking...  I'll raise at next DevCoven I think, and if we feel we should use it for something like this, I'll raise a TechDebt story to fix it?

davidgouge: Ok, no probs, just a thought.  Not a deal breaker. :)

PR #437, opened 2018-10-02
ON-353 - Staff UI Reports

martin-hornsby:


src/SFA.DAS.AssessorService.Application.Api/Controllers/Staff/StaffReportsController.cs
```diff
@@ -0,0 +1,66 @@
...
+using Swashbuckle.AspNetCore.SwaggerGen;
+using System;
+using System.Collections.Generic;
+using System.Data.SqlClient;
+using System.Net;
+using System.Threading.Tasks;
+
+namespace SFA.DAS.AssessorService.Application.Api.Controllers.Staff
+{
+    [Authorize(Roles = "AssessorServiceInternalAPI")]
```
davidgouge: This will need to just be `[Authorize]`  There are no roles in the API.

src/SFA.DAS.AssessorService.Application.Api/Controllers/Staff/StaffReportsController.cs
```diff
@@ -0,0 +1,66 @@
...
+            _logger = logger;
+        }
+
+        [HttpGet()]
+        [SwaggerResponse((int)HttpStatusCode.OK, Type = typeof(IEnumerable<StaffReport>))]
+        [SwaggerResponse((int)HttpStatusCode.BadRequest, typeof(IDictionary<string, string>))]
+        [SwaggerResponse((int)HttpStatusCode.InternalServerError, Type = typeof(ApiResponse))]
+        public async Task<IActionResult> GetReportList()
+        {
+            _logger.LogInformation($"Recieved request to get list of reports");
```
davidgouge: 🎶 "I before E, except after C" (And except all of the other exceptions 😆  )

PR #438, opened 2018-10-02
On 891 add new organisation

MarkFCain:


src/SFA.DAS.AssessorService.Data/RegisterQueryRepository.cs
```diff
@@ -152,6 +152,28 @@ public async Task<bool> EpaOrganisationStandardExists(string organisationId, int
...
+                    await connection.OpenAsync();
+                
+                var sqlToCheckExists =
+                    "select CASE count(0) WHEN 0 THEN 0 else 1 end result FROM [Organisations] " +
+                    $@"WHERE EndPointAssessorName = '{organisationName}'";
+                
+                if (!string.IsNullOrEmpty(organisationIdToExclude))
+                {
+                    sqlToCheckExists =  "select CASE count(0) WHEN 0 THEN 0 else 1 end result FROM [Organisations] " +
+                                        $@"WHERE EndPointAssessorName = '{organisationName}' AND  EndPointAssessorOrganisationId != '{organisationIdToExclude}'";
```
davidgouge: Can you parameterise these string interpolations?  I'm not sure whether Sql Injection can happen with `{organisationName}` but better safe than sorry I guess. 😀

src/SFA.DAS.AssessorService.Application.Api.UnitTests/Validators/Register/EpaOrganisationValidatorTests.cs
```diff
@@ -207,5 +210,76 @@ public void CheckIfOrganisationStandardHasValidContactIdReturnsAnErrorMessage(st
             else
             _registerRepository.Verify(r => r.ContactIdIsValidForOrganisationId(It.IsAny<string>(), It.IsAny<string>()), Times.Once);
```
jcoxhead: missing indentation? very picky point :)

src/SFA.DAS.AssessorService.Application.Api/Validators/EpaOrganisationValidator.cs
```diff
@@ -65,6 +72,20 @@ public string CheckUkprnIsValid(long? ukprn)
...
+                string.Empty;
+        }
+
+        public string CheckOrganisationNameNotUsedForOtherOrganisations(string name, string organisationIdToIgnore)
+        {
+            return _registerRepository.EpaOrganisationAlreadyUsingName(name, organisationIdToIgnore).Result 
+                ? ErrorMessageOrganisationNameAlreadyPresent : 
+                string.Empty;
+        }
+
```
jcoxhead: Would if possible consider using await ? think result is synchronous - see https://stackoverflow.com/questions/24623120/await-on-a-completed-task-same-as-task-result

PR #440, opened 2018-10-03
ON-970 added code and page to run import

MarkFCain:
Dirty, dirty code to add a page to be able to 1mp0rt organisation details from a spreadsheet (to call the patch endpoint that runs this).  Please just wave through, it's too painful to look at.


src/SFA.DAS.AssessorService.Web.Staff/Views/Register/ImPage.cshtml
```diff
@@ -0,0 +1,18 @@
...
+@{
+    ViewBag.Title = "ImPage";
+    Layout = "_Layout";
+}
+
+
+
+<br/>
+<br/>
+<br/>
```
davidgouge: Love it! 😆

PR #441, opened 2018-10-08
On 893 edit organisation

MarkFCain:
Hi guys, this seems like a lot of changes, but it actually contains not only the code to edit an existing organisation, I also changed the codebase to not rely on our own versions of WIN class types, but actually use the types directly from WIN API.Types.

I've removed all the objects that were in the
using SFA.DAS.AssessorService.ExternalApis.AssessmentOrgs.Types;

And updated references to them with
using SFA.DAS.Apprenticeships.Api.Types

The types updated are:
Address
Link
Organisation
OrganisationSummary
Period
Provider
Standard
StandardOrganisationSummary

They are mostly the same, but significantly our local version of 'Organisation' had an extra field 'ukprn'.  I couldn't find anywhere this was used or hydrated (Win wouldn't be passing it back)... but be aware of that when you code review.   

If you know these types are in anything you're currently working on, it might be worth subbranching that, and pull this branch in, and making sure nothing breaks... Just a suggestion....


src/SFA.DAS.AssessorService.Application.UnitTests/Handlers/RegisterUpdate/WhenRegisterUpdateCalled_AndItReturnsUnknownEpao.cs
```diff
@@ -26,13 +26,13 @@ public void Arrange()
                     new OrganisationSummary {Id = "EPA0003", Ukprn = 33333333}
                 }.AsEnumerable()));
 
-            ApiClient.Setup(c => c.Get("EPA0003")).Returns(new ExternalApis.AssessmentOrgs.Types.Organisation { Id = "EPA0003", Name = "A New EPAO", UkPrn = 33333333});
+            ApiClient.Setup(c => c.Get("EPA0003")).Returns(new Organisation { Id = "EPA0003", Name = "A New EPAO"});
```
jcoxhead: Don;t know if it helps but I am using  <PackageReference Include="RichardSzalay.MockHttp" Version="4.0.0" />

there are examples in SFA.DAS.AssessorService.Web.Staff.Tests.MockedObjects:public class MockedApiClient

might be worthwhile  a look.

src/SFA.DAS.AssessorService.Application.UnitTests/Handlers/Search/When_returned_learners_match_users_epaorgid.cs
```diff
@@ -5,12 +5,13 @@
...
 using NUnit.Framework;
+using SFA.DAS.Apprenticeships.Api.Types;
+using SFA.DAS.Apprenticeships.Api.Types.AssessmentOrgs;
 using SFA.DAS.AssessorService.Api.Types.Models;
 using SFA.DAS.AssessorService.Application.Handlers.Search;
 using SFA.DAS.AssessorService.Application.Interfaces;
 using SFA.DAS.AssessorService.Domain.Entities;
 using SFA.DAS.AssessorService.ExternalApis.AssessmentOrgs;
-using SFA.DAS.AssessorService.ExternalApis.AssessmentOrgs.Types;
 using Organisation = SFA.DAS.AssessorService.Domain.Entities.Organisation;
```
jcoxhead: Picky comment but is alias required?

src/SFA.DAS.AssessorService.Application/Handlers/ao/GetStandardsByAssessmentOrganisationHandler.cs
```diff
@@ -28,13 +28,7 @@ public async Task<List<OrganisationStandardSummary>> Handle(GetStandardsByOrgani
             var organisationId = request.OrganisationId;
             _logger.LogInformation($@"Handling OrganisationStandards Request for OrganisationId [{organisationId}]");
             var orgStandards = await _registerQueryRepository.GetOrganisationStandardByOrganisationId(organisationId);
```
jcoxhead: use of orgStandards instead of organisationStandards?

PR #447, opened 2018-10-15
On 897 add contact front end

MarkFCain:
Hi chaps, this is a compare based on the approved ON-959. The front end for an 'add contact'.


MarkFCain:
Will be merged as part of ON-966 merge

PR #448, opened 2018-10-15
On 947 staff ui phase2

martin-hornsby:


src/SFA.DAS.AssessorService.Web.Staff/Views/CertificateAmend/Check.cshtml
```diff
@@ -10,9 +11,9 @@
         @{
             var fullName = @Model.GivenNames + " " + @Model.FamilyName;
         }
-  
+
         <a class="link-back" asp-controller="Search" asp-action="Select" asp-route-stdcode=@Model.StandardCode
-           asp-route-uln=@Model.Uln asp-route-searchstring=@fullName >Back</a>
+           asp-route-uln=@Model.Uln asp-route-searchstring=@fullName>Back</a>
```
MarkFCain: This looks wrong, but I leave it you your judgement
asp-route-searchstring=@fullName

Could fullname have spaces eg 'joe cool'?  In which case maybe asp-route-searchstring="@fullName" might be better?

PR #449, opened 2018-10-15
On 961 api put existing contact

MarkFCain:
Another PR against the branch before.


src/SFA.DAS.AssessorService.Api.Types/Models/Register/UpdateEpaOrganisationContactRequest.cs
```diff
@@ -0,0 +1,12 @@
+﻿using MediatR;
+
+namespace SFA.DAS.AssessorService.Api.Types.Models.Register
+{
+    public class UpdateEpaOrganisationContactRequest : IRequest<string>
```
martin-hornsby: Just wondering if this be better if it returned a Contact object or an UpdateEpaOrganisationContactResponse than a string

src/SFA.DAS.AssessorService.Data/RegisterValidationRepository.cs
```diff
@@ -143,5 +143,31 @@ public async Task<bool> EmailAlreadyPresentInAnotherOrganisation(string email, s
...
+
+        public async Task<bool> ContactExists(string contactId)
+        {
+            using (var connection = new SqlConnection(_configuration.SqlConnectionString))
+            {
+                if (connection.State != ConnectionState.Open)
+                    await connection.OpenAsync();
+                const string sqlToCheckExists =
+                    "select CASE count(0) WHEN 0 THEN 0 else 1 end result FROM [Contacts] " +
+                    "WHERE convert(varchar(50),id)  = @contactId";
```
martin-hornsby: Could this be done using GUID than string?

src/SFA.DAS.AssessorService.Api.Types/Models/Register/UpdateEpaOrganisationContactRequest.cs
MarkFCain: I don't need the contact returned, just the id of the contact to work with the MVC stateless flow.  There's a separate endpoint to get the contact from the id returned, if needed.

MarkFCain:
Closed as will be merged as part of ON-966

PR #458, opened 2018-10-22
Fixed bug with Learner Name capitalisation.  Added tests. Consolidated Test projects.

davidgouge:


src/SFA.DAS.AssessorService.EpaoImporter/DomainServices/PrintingSpreadsheetCreator.cs
```diff
@@ -163,40 +163,38 @@ private static void CreateWorksheetHeader(int batchNumber, ExcelWorksheet worksh
...
-            for (var i = 0; i + 1 < chars.Length; i++)
+        private static void FixTitleCaseForNamesWithApostrophesAndHyphens(char[] newNameCharacters)
+        {
+            for (var i = 0; i + 1 < newNameCharacters.Length; i++)
             {
-                if (chars[i].Equals('\'') ||
-                    chars[i].Equals('`') ||
-                    chars[i].Equals('-'))
-                {                    
-                    chars[i + 1] = char.ToUpper(chars[i + 1]);
```
MarkFCain: This is probably extra fussy, but this will break if any of those symbols (quote, hyphen) appear as last character in string.  I'm not going to no approve based on that, but as you're closer to domain issue, please evaluate and fix if needed.

davidgouge: Bloody good point.  👍

PR #460, opened 2018-10-22
On 962 update existing organisation standard endpoints

MarkFCain:
Compare to ON-966 as there's some delay mergining ON-966 with master caused by GitHub problems


src/SFA.DAS.AssessorService.Application/Handlers/EpaOrganisationHandlers/CreateEpaOrganisationStandardHandler.cs
```diff
@@ -30,30 +32,34 @@ public CreateEpaOrganisationStandardHandler(IRegisterRepository registerReposito
...
-            ThrowNotFoundExceptionIfErrorPresent(errorDetails);
+            var validationResponse = _validator.ValidatorCreateEpaOrganisationStandardRequest(request);
 
-            errorDetails.Append(_validator.CheckIfStandardNotFound(request.StandardCode));
-            ThrowNotFoundExceptionIfErrorPresent(errorDetails);
+            if (!validationResponse.IsValid)
+            {
+                var message = validationResponse.Errors.Aggregate(string.Empty, (current, error) => current + error.ErrorMessage + "; ");
+                _logger.LogError(message);
+                if (validationResponse.Errors.Any(x => x.StatusCode == ValidationStatusCode.BadRequest.ToString()))
```
martin-hornsby: Is there any chance that ValidationResponse.StatusCode could be the same type as ValidationStatusCode rather than doing string comparison?

PR #461, opened 2018-10-23
On 692 privately funded apprentice certificate requests digital solution

jcoxhead:


src/SFA.DAS.AssessorService.Application/Handlers/Staff/StaffSearchHandler.cs
```diff
@@ -35,39 +39,56 @@ public class StaffSearchHandler : IRequestHandler<StaffSearchRequest, PaginatedL
...
             if (regex.IsMatch(request.SearchQuery))
-            {
-                return await _staffIlrRepository.SearchForLearnerByEpaOrgId(request);
+            {                
+                var sr = await _staffIlrRepository.SearchForLearnerByEpaOrgId(request);
+                sr.DisplayEpao = true;
+                return sr;
             }
 
             if (request.SearchQuery.Length == 10 && long.TryParse(request.SearchQuery, out var uln))
```
MarkFCain: Don't know this domain, so this is just a query that can be ignored.  Is length 10 and only 10 right here?

src/SFA.DAS.AssessorService.Data/CertificateRepository.cs
```diff
@@ -38,9 +38,49 @@ public async Task<Certificate> New(Certificate certificate)
...
+            _context.Certificates.Add(certificate);
+            try
+            {
+                _context.SaveChanges();
+            }
+            catch (Exception e)
+            {
+                if (!(e.InnerException is SqlException sqlException)) throw;
+
+                if (sqlException.Number == 2601 || sqlException.Number == 2627)
```
MarkFCain: I'm really not a fan of exception handling as flow, but I understand the pressures and don't want to drag.  Can a tech debt story be raised to precheck what causes these 2 error codes and we finesse out the exception as program stuff here?  I suggest a short spike story to investigate and establish under tech debt, put straight into the backlog for a less time-constrained point in the future.

src/SFA.DAS.AssessorService.Data/CertificateRepository.cs
```diff
@@ -53,11 +93,13 @@ public async Task<Certificate> New(Certificate certificate)
                 if (sqlException.Number == 2601 || sqlException.Number == 2627)
                 {
```
MarkFCain: My previous comment about exception as flow is true of this too. Again, I vote for don't hold up this story, put this and the other one it a little spike investigation for a future date.

src/SFA.DAS.AssessorService.EpaoImporter/Data/AssessorServiceApi.cs
```diff
@@ -49,6 +49,30 @@ public async Task<IEnumerable<CertificateResponse>> GetCertificatesToBePrinted()
...
+            var certificates = await response.Content.ReadAsAsync<List<CertificateResponse>>();
+            if (response.IsSuccessStatusCode)
+            {
+                _aggregateLogger.LogInfo($"Getting Certificates to be printed - Status code returned: {response.StatusCode}. Content: {response.Content.ReadAsStringAsync().Result}");
+            }
+            else
+            {
+                _aggregateLogger.LogInfo($"Getting Certificates to be printed - Status code returned: {response.StatusCode}. Content: {response.Content.ReadAsStringAsync().Result}");
+            }
+
```
MarkFCain: The If statement is redundant, both if and else are same

src/SFA.DAS.AssessorService.Data/CertificateRepository.cs
davidgouge: These are specifically checked for Unique Index Constraint violations as even with some very close checking for existence before doing an insert, we were still getting the odd ninja getting through.

MarkFCain: If its there to be belt when there's already braces applied, then it's fine as is, nothing else to do.  (I hope I'm not straining the metaphor too much there...)

MarkFCain: (though I'd maybe refactor out into a local method called LastStandMeasureToStopNinjaI|nsertViolationWeAreAlreadyGuardingAgainst...)

PR #466, opened 2018-10-29
ON-981 - Added OrganisationData to AssessmentOrganisationSummary as A…

martin-hornsby:
Apply service require Organisation Data when Searching

### DIFF
--- a/src/SFA.DAS.AssessorService.Data/RegisterQueryRepository.cs
+++ b/src/SFA.DAS.AssessorService.Data/RegisterQueryRepository.cs
@@ -217,7 +217,7 @@ namespace SFA.DAS.AssessorService.Data
                     await connection.OpenAsync();
 
                 var sqlForStandardByOrganisationId =
-                     "SELECT distinct EndPointAssessorOrganisationId as organisationId, StandardCode, EffectiveFrom, EffectiveTo, DateStandardApprovedOnRegister, ContactId, OrganisationData " +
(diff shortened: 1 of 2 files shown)


src/SFA.DAS.AssessorService.Data/RegisterQueryRepository.cs
```diff
@@ -217,7 +217,7 @@ public async Task<IEnumerable<OrganisationStandardSummary>> GetOrganisationStand
                     await connection.OpenAsync();
 
                 var sqlForStandardByOrganisationId =
-                    "SELECT distinct EndPointAssessorOrganisationId as organisationId, StandardCode, EffectiveFrom, EffectiveTo, DateStandardApprovedOnRegister, ContactId "+
+                     "SELECT distinct EndPointAssessorOrganisationId as organisationId, StandardCode, EffectiveFrom, EffectiveTo, DateStandardApprovedOnRegister, ContactId, OrganisationData " +
                      "FROM [OrganisationStandard] WHERE EndPointAssessorOrganisationId = @organisationId";
                 return await connection.QueryAsync<OrganisationStandardSummary>(sqlForStandardByOrganisationId, new {organisationId});
```
MarkFCain: This object 'OrganisationStandardSummary' doesn't have a field for OrganisationData.  Either add it to that class, or remove it from this sql.  I favour the first unless you have a very good reason to add it.

martin-hornsby: Agreed, it was superfluous so I removed it

PR #467, opened 2018-10-30
DASD-2906 ARM Template Updates

qamago:


EwanNoble:
This is missing Application Insights resources for each of the app services, these need to then be referenced in the AppSettings

PR #468, opened 2018-10-30
On 971 show manual certificates in previous history tab and staff app

jcoxhead:


src/SFA.DAS.AssessorService.Application/Handlers/Certificates/GetPrivateCertificatePrivateByUlnHandler.cs
```diff
@@ -0,0 +1,28 @@
...
+        public GetPrivateCertificatePrivateByUlnHandler(ICertificateRepository certificateRepository)
+        {
+            _certificateRepository = certificateRepository;
+        }
+
+        public async Task<Certificate> Handle(GetPrivateCertificatePrivateByUlnRequest request,
+            CancellationToken cancellationToken)
+        {
+            var certificate = await _certificateRepository.GetPrivateCertificate(request.Uln, request.EndPointAssessorOrganisationId, request.LastName);
+//            var certificateResponse = Mapper.Map<Certificate, CertificateResponse>(certificate);
```
MarkFCain: unneeded comment

src/SFA.DAS.AssessorService.Database/Tables/Certificates.sql
```diff
@@ -17,6 +17,7 @@
     [ProviderUkPrn] INT NOT NULL, 
     [CertificateReferenceId] INT NOT NULL IDENTITY(10001,1), 
 	[LearnRefNumber] NVARCHAR(12) NULL,
+	[ReasonForRejection] [nvarchar](max) NOT NULL,
```
MarkFCain: check you need this

src/SFA.DAS.AssessorService.Application/Handlers/Staff/LearnerDetailHandler.cs
```diff
@@ -52,12 +64,12 @@ public async Task<LearnerDetail> Handle(LearnerDetailRequest request, Cancellati
...
                 }
-                
+
 
                 certificateData = JsonConvert.DeserializeObject<CertificateData>(certificate.CertificateData);
                 epao = await _organisationRepository.Get(certificate.OrganisationId);
             }
-            
+
             var learnerDetail = new LearnerDetail()
```
MarkFCain: Just a suggestion.  I would put this into a private mapper method to eliminate the bloat inside this function.  Your choice though.

src/SFA.DAS.AssessorService.Database/Tables/Certificates.sql
jcoxhead: agreed tx will fix now

src/SFA.DAS.AssessorService.Application/Handlers/Staff/LearnerDetailHandler.cs
jcoxhead: yep hang over from past from daves code will see if I get time to resolve for tommorrow.

src/SFA.DAS.AssessorService.Api.Types/Models/Certificates/GetApprovedDraftCertificatesRequest.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Collections.Generic;
+using MediatR;
+
+namespace SFA.DAS.AssessorService.Api.Types.Models.Certificates
+{
+    public class GetApprovedDraftCertificatesRequest : IRequest<List<CertificateSummaryResponse>>
+    {
```
martin-hornsby: Is this empty by design? If so can you pop a quick comment in here.

src/SFA.DAS.AssessorService.Application.Api/Controllers/CertificateController.cs
```diff
@@ -107,5 +107,15 @@ public async Task<IActionResult> Approvals([FromBody] CertificateApprovalRequest
...
             return Ok();
         }
+
+        [HttpPost("prepareprivatecertificatesforbatchrun", Name = "PreparePrivateCertificatesforBatchRun")]
+        [SwaggerResponse((int)HttpStatusCode.OK, Type = typeof(Certificate))]
+        [SwaggerResponse((int)HttpStatusCode.BadRequest, typeof(IDictionary<string, string>))]
+        [SwaggerResponse((int)HttpStatusCode.InternalServerError, Type = typeof(ApiResponse))]
+        public async Task<IActionResult> PreparePrivateCertificatesforBatchRun([FromBody] PrivateCertificatePrepareForBatchRunRequest privateCertificatePrepareForBatchRunRequest)
+        {
+            await _mediator.Send(privateCertificatePrepareForBatchRunRequest);
```
martin-hornsby: The await is outside of the return OK() - is this by design?

src/SFA.DAS.AssessorService.Application/Handlers/Certificates/PrivateCertificatePrepareForBatchRunHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+            var certificates = await _certificateRepository.GetCertificates(statuses);
+            foreach (var certificate in certificates)
+            {
+                var certificateData = JsonConvert.DeserializeObject<CertificateData>(certificate.CertificateData);
+                if (certificate.Status == Domain.Consts.CertificateStatus.Approved)
+                {
+                    certificate.Status = Domain.Consts.CertificateStatus.Submitted;
+                }
+
+                if (certificate.Status == Domain.Consts.CertificateStatus.Rejected)
```
martin-hornsby: Whilst not a issue right now. Think this should be an else if block

src/SFA.DAS.AssessorService.Api.Types/Models/Certificates/GetApprovedDraftCertificatesRequest.cs
jcoxhead: done

src/SFA.DAS.AssessorService.Application.Api/Controllers/CertificateController.cs
jcoxhead: yep there's no response with it.

src/SFA.DAS.AssessorService.Application/Handlers/Certificates/PrivateCertificatePrepareForBatchRunHandler.cs
jcoxhead: in this case I would have to leave a default case which is more code, so have not done this.

alandburns:
Reworked elsewhere

PR #470, opened 2018-11-01
ON-1017 - Retrieve Organisation Type as part of EPAO Register search

martin-hornsby:


src/SFA.DAS.AssessorService.Data/RegisterQueryRepository.cs
```diff
@@ -263,7 +265,13 @@ public async Task<IEnumerable<AssessmentOrganisationSummary>> GetAssessmentOrgan
             {
                 if (connection.State != ConnectionState.Open)
                     await connection.OpenAsync();
-                var assessmentOrganisationSummaries = await connection.QueryAsync<AssessmentOrganisationSummary>("select EndPointAssessorOrganisationId as Id, EndPointAssessorName as Name, EndPointAssessorUkprn as ukprn, OrganisationData from [Organisations] where EndPointAssessorOrganisationId like @organisationId", new {organisationId = $"{organisationId.Replace(" ","")}%" });
+
+                var sql =
+                      "SELECT o.EndPointAssessorOrganisationId as Id, o.EndPointAssessorName as Name, o.EndPointAssessorUkprn as ukprn, o.OrganisationData, ot.Id as OrganisationTypeId, ot.Type as OrganisationType "
+                    + "FROM [Organisations] o LEFT JOIN [OrganisationType] ot ON ot.Id = o.OrganisationTypeId "
```
MarkFCain: Im a little rusty on this, but I think this should be an outer left join, just in case.  OrganisationTypeId is a non mandatory field on organisations

src/SFA.DAS.AssessorService.Data/RegisterQueryRepository.cs
```diff
@@ -274,7 +282,13 @@ public async Task<IEnumerable<AssessmentOrganisationSummary>> GetAssessmentOrgan
             {
                 if (connection.State != ConnectionState.Open)
                     await connection.OpenAsync();
-                var assessmentOrganisationSummaries = await connection.QueryAsync<AssessmentOrganisationSummary>("select EndPointAssessorOrganisationId as Id, EndPointAssessorName as Name, EndPointAssessorUkprn as ukprn, OrganisationData from [Organisations] where replace(EndPointAssessorName, ' ','') like @organisationName", new {organisationName =$"%{organisationName.Replace(" ","")}%" } );
+
+                var sql =
+                      "SELECT o.EndPointAssessorOrganisationId as Id, o.EndPointAssessorName as Name, o.EndPointAssessorUkprn as ukprn, o.OrganisationData, ot.Id as OrganisationTypeId, ot.Type as OrganisationType "
+                    + "FROM [Organisations] o LEFT JOIN [OrganisationType] ot ON ot.Id = o.OrganisationTypeId "
```
MarkFCain: OUTER LEFT JOIN imo

PR #474, opened 2018-11-05
On 963 edit organisation standard front end

MarkFCain:
It looks like a lot, but most is config and front end page stuff


MarkFCain:
merged in as part of ON-1052

PR #480, opened 2018-11-07
On 895 make org standard live

MarkFCain:


src/SFA.DAS.AssessorService.Web.Staff/Views/Register/EditOrganisationStandard.cshtml
```diff
@@ -203,8 +204,12 @@
                 <textarea class="form-control form-control-3-4" maxlength="500" name="Comments" id="Comments" sfa-validationerror-class="form-control-error" asp-for="Comments" rows="5"></textarea>
             </div>
             <div class="form-group">
-                <button class="button" type="submit">Save</button>
-                <a class="button text-link" href="/register/view-standard/@Model.OrganisationStandardId">Cancel</a>
+                <button class="button" type="submit" name="ActionChoice" value="Save">Save</button>
+                @if (Model.OrganisationStatus == "Live" && Model.Status == "New")
```
martin-hornsby: Might be worth putting these into defined constants rather than magic strings

MarkFCain: Will log as Tech Debt

MarkFCain:
Merged in as part of ON-1052

PR #486, opened 2018-11-12
ON-1126 - Added Primary Contact email on EPAO Register - Search Organisation

martin-hornsby:
Apply Service needs to alert the primary contact of the organisation. Thus we need to expose email when searching the EPAO Register.


martin-hornsby:
@alandburns I have done the requested change. Could you re-review? thanks :)

PR #505, opened 2018-11-24
On 1185 search hotfix

martin-hornsby:


src/SFA.DAS.AssessorService.Application/Handlers/Search/SearchResultExtensions.cs
```diff
@@ -54,7 +54,7 @@ public static List<SearchResult> MatchUpExistingCompletedStandards(this List<Sea
                 if (submittedLogEntry == null) continue;
 
                 var submittingContact = contactRepository.GetContact(submittedLogEntry.Username).Result ?? contactRepository.GetContact(certificate.UpdatedBy).Result;
-                var createdContact = contactRepository.GetContact(createdLogEntry.Username).Result ?? contactRepository.GetContact(certificate.CreatedBy).Result;
+                var createdContact = contactRepository.GetContact(createdLogEntry?.Username).Result ?? contactRepository.GetContact(certificate.CreatedBy).Result;
```
MarkFCain: Can I just check that passing a null value for 'createdLogEntry' into GetContact returns a null as the result? This doesn't fall over? 

If that's all ok, then I approve

alandburns: For now we will proceed with this fix.
But need to look again at the logic used to decide who can see what level of Certificate data.

PR #507, opened 2018-11-27
Created a string extension class which corrects for learner namecases…

arshedmust:
… using various regular expressions. Updated unit tests accordingly. Finally removed the current mechanism to cater for namecases.


src/SFA.DAS.AssessorService.Data/AssessmentOrgsImporter.cs
```diff
@@ -216,7 +216,7 @@ public BlobContainerHelper(IWebConfiguration webConfiguration)
         }
         public async Task<CloudBlobContainer> GetContainer(string containerName)
         {
-            var storageAccount = CloudStorageAccount.Parse(_webConfiguration.IFATemplateStorageConnectionString);     
+            var storageAccount = CloudStorageAccount.DevelopmentStorageAccount;
```
davidgouge: Should this change be pushed? 😸

arshedmust: As far as I understand this bit of functionality will only be used to import data to local dev database from a spreadsheet for dev use.

PR #534, opened 2018-12-13
On 1281 new fields in organisations

MarkFCain:


src/SFA.DAS.AssessorService.Application/Handlers/EpaOrganisationHandlers/CreateEpaOrganisationHandler.cs
```diff
@@ -71,7 +71,8 @@ private static EpaOrganisation MapOrganisationRequestToOrganisation(CreateEpaOrg
                     Address4 = request.Address4,
                     LegalName = request.LegalName,
                     Postcode = request.Postcode,
-                    WebsiteLink = request.WebsiteLink
+                    WebsiteLink = request.WebsiteLink,
+
```
martin-hornsby: Is this intentional not to include  CompanyNumber & CharityNumber

MarkFCain: good point, now fixed

PR #543, opened 2019-01-11
Apply staff feedback

alandburns:
Get up to date with master


martin-hornsby:
This is not ready for merging back into master at this moment in time

src/SFA.DAS.AssessorService.Web.Staff/Controllers/Apply/FinancialController.cs
```diff
@@ -0,0 +1,200 @@
...
+                return File(compressedBytes, "application/zip", "FinancialDocuments.zip");
+            }
+            
+            
+            
+
+            
+//            var stream = await downloadedFile.Content.ReadAsStreamAsync();
+//            var contentType = downloadedFile.Content.Headers.ContentType.ToString();
+        }
```
MarkFCain: lot of space, remove the comments?

davidgouge:
> Nothing major, except just wanted to check two things: (A) are we intending to create a common types library maybe for these two repos (in one or the other) at some future date? Lot of duplication of types in both now. (2) is all the css updates correct? Losing/changing css stuff makes me nervous as it can be a last-minute wrecking ball on a site's appearance.

(A) Yes.  That would have been a great thing to create at the start.  I'm sure we'll find time to refactor one in.
(2) No idea.

PR #549, opened 2019-01-23
Add/Edit companies house and charity number

David-B-Read:


src/SFA.DAS.AssessorService.Application.Api/Validators/EpaOrganisationValidator.cs
```diff
@@ -28,7 +28,10 @@ public class EpaOrganisationValidator: IEpaOrganisationValidator
         private readonly IRegisterQueryRepository _registerQueryRepository;
         private readonly IStringLocalizer<EpaOrganisationValidator> _localizer;
         private readonly ISpecialCharacterCleanserService _cleanserService;
-       
+
+        private const string CompaniesHouseNumberRegex = "[A-Z0-9]{2}[0-9]{6}";
```
MarkFCain: Consider whether this needs to allow lower case characters in the first 2 values.  And then convert the save point to uppercase?  Makes it less nitty for users.
I'll leave that to your discretion, not going to not approve because of this.

src/SFA.DAS.AssessorService.Web.Staff/Views/Register/EditOrganisation.cshtml
```diff
@@ -72,6 +72,20 @@ Layout = "_Layout";
                     </label>
                     <input class="form-control" name="LegalName" sfa-validation-for="LegalName" sfa-validationerror-class="form-control-error" asp-for="LegalName" type="text" />
                 </div>
+                <div class="form-group" sfa-validation-for="CompanyNumber" sfa-validationerror-class="form-group-error">
+                    <label for="CompanyNumber">
+                        <span class="form-label-bold">Company number (optional)</span>
```
MarkFCain: We haven't used the phrase '(optional)' anywhere else in this app, so consider removing it in both places? 
I'll leave that to your discretion, I won't make not approve because of this.

src/SFA.DAS.AssessorService.Application.Api/Validators/EpaOrganisationValidator.cs
MarkFCain: Sounds good.  Probably good to move to testing once done then.

PR #553, opened 2019-01-30
On 1058 fha imports and endpoint changes

MarkFCain:
Note the copy of deleted code from postdeploy into postbuild import script 💃


src/SFA.DAS.AssessorService.Api.Types/Models/AO/OrganisationData.cs
```diff
@@ -11,5 +13,7 @@ public class OrganisationData
         public string Postcode { get; set; }
         public string CompanyNumber { get; set; }
         public string CharityNumber { get; set; }
+        public DateTime? FinancialDueDate { get; set; }
+        public bool? FinancialExempt { get; set; }
```
martin-hornsby: nullable bool is a bit smelly for what is effectively a yes/no answer.

MarkFCain: it's a yes/no/don't know answer, so valid here

PR #557, opened 2019-01-31
ON-1446 main coding done

MarkFCain:


MarkFCain:
This includes the code from story ON-1058, combined here into one merge

PR #562, opened 2019-02-01
On 1435 ilrs data synch stage 2

arshedmust:


src/SFA.DAS.AssessorService.EpaoDataSync/RefreshIlrsHttpFuncionApp.cs
```diff
@@ -11,26 +11,26 @@
 
 namespace SFA.DAS.AssessorService.EpaoDataSync
 {
-    public static class RefreshIlrsFuncionApp
+    public static class RefreshIlrsHttpFuncionApp
```
davidgouge: Can you fix the typo in this class please? :)

arshedmust: Corrected

PR #565, opened 2019-02-01
On 1436 validation api endpoints

MarkFCain:


MarkFCain:
closed because branched from something not yet committed so wanted to keep this locked down till that is in branch

PR #568, opened 2019-02-04
On 1436 validation api endpoints

MarkFCain:


src/SFA.DAS.AssessorService.Api.Types/Models/Validation/ValidationRequest.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace SFA.DAS.AssessorService.Api.Types.Models.Validation
+{
+    public class ValidationRequest : IRequest<bool>
+    {
+
+
+        public string ValidationType { get; set; }
+        public string ValidationString { get; set; }
+        public string ValidationMatchValue { get; set; }
```
davidgouge: Maybe 

`public string Type { get; set; }

        public string String { get; set; }

        public string MatchValue { get; set; }`  

we know they are Validation related. 👍 ❓

MarkFCain: sounds legit.  Not calling a field 'String' though!  I Think ValueToCheck maybe...

PR #571, opened 2019-02-06
On 1481 return to dashboard links

MarkFCain:


martin-hornsby:
we were having merge issues. this branch is to be ignored

PR #585, opened 2019-02-18
On 1506 add new organisation and contact

MarkFCain:


alandburns:
Work in progress. Merge now as the logic has been tested.

PR #589, opened 2019-02-20
Fix Export

alandburns:


src/SFA.DAS.AssessorService.Database/StoredProcedures/EPAO_Register_register_organisation.sql
```diff
@@ -15,8 +15,14 @@ select EndPointAssessorOrganisationId as EPA_organisation_identifier,
...
 	JSON_VALUE(OrganisationData,'$.LegalName') as 'Legal Name'
-	 from organisations o left outer join organisationType ot on o.OrganisationTypeId = ot.Id
+	 from organisations o 
+	 left outer join organisationType ot on o.OrganisationTypeId = ot.Id
+	 join (select EndPointAssessorOrganisationId 
+		   from OrganisationStandard
+		   where Status <> 'Deleted'
+		   group by EndPointAssessorOrganisationId
+     ) ab ON ab.EndPointAssessorOrganisationId = o.EndPointAssessorOrganisationId
 	 WHERE o.EndPointAssessorOrganisationId<> 'EPA0000'
```
David-B-Read: WHERE o.EndPointAssessorOrganisationId<> -space needed here?

alandburns: It does work .. but good point

PR #590, opened 2019-02-20
On 1500 use standard collation for standard name

MarkFCain:


src/SFA.DAS.AssessorService.Application.Api/Validators/EpaOrganisationValidator.cs
```diff
@@ -185,17 +188,7 @@ public string CheckIfOrganisationNotFound(string organisationId)
 
         public Standard GetStandard(int standardCode)
```
martin-hornsby: Can this be made async?

src/SFA.DAS.AssessorService.Application/Handlers/Staff/StaffSearchHandler.cs
```diff
@@ -142,16 +143,16 @@ private List<StaffSearchItems> MatchUpExistingCompletedStandards(List<StaffSearc
             return searchResults;
         }
 
-        private List<StaffSearchItems> PopulateStandards(List<StaffSearchItems> searchResults, IAssessmentOrgsApiClient assessmentOrgsApiClient, ILogger<SearchHandler> logger)
+        private List<StaffSearchItems> PopulateStandards(List<StaffSearchItems> searchResults, IStandardService standardService, ILogger<SearchHandler> logger)
```
martin-hornsby: Could this be made async?

MarkFCain: yeah, I can do that, and will do.  But this isn't new code, I've just changed the source of data :D

src/SFA.DAS.AssessorService.Application/Handlers/Staff/StaffSearchHandler.cs
```diff
@@ -142,16 +143,16 @@ private List<StaffSearchItems> MatchUpExistingCompletedStandards(List<StaffSearc
...
         {
-            var allStandards = assessmentOrgsApiClient.GetAllStandards().Result;
+            var allStandards = standardService.GetAllStandards().Result;
 
             foreach (var searchResult in searchResults)
             {
                 var standard = allStandards.SingleOrDefault(s => s.StandardId == searchResult.StandardCode.ToString());
```
martin-hornsby: Is there scope to change StandardId to an int type?

MarkFCain: I tried to do this before, but after about 2 days, gave up and had to reel it back.  It's all over the place.

This is work  best done (if at all) when we go through the second phase of making standardCollation the only source of truth about standards inside assessor-service.  As all objects will be mapped into that.

martin-hornsby: Fair comment :)

PR #592, opened 2019-02-21
On 1389 staff app search performance

alandburns:


alandburns:
Nasty query fix to get either Ilrs records or certificates in the Staff Search.

src/SFA.DAS.AssessorService.Application/Handlers/Staff/StaffSearchHandler.cs
```diff
@@ -96,9 +96,9 @@ private async Task<StaffReposSearchResult> Search(StaffSearchRequest request)
                 // Search string is a long of 10 length so must be a uln.
                 var sr = new StaffReposSearchResult
```
martin-hornsby: This code works, albeit is a bit smelly. Just wondering if this looks cleaner:

var searchResults = (await _staffIlrRepository.SearchForLearnerByUln(request)).ToList();
return new StaffReposSearchResult
                {
                    PageOfResults = searchResults,  
                    TotalCount = searchResults .Count
                 };

martin-hornsby: This code works, albeit is a bit smelly. Just wondering if this looks cleaner:

var results = await _staffIlrRepository.SearchForLearnerByUln(request);

return new StaffReposSearchResult
{
	PageOfResults = results,
	TotalCount = results.Count()
};

src/SFA.DAS.AssessorService.Application/Handlers/Staff/StaffSearchHandler.cs
```diff
@@ -107,9 +107,9 @@ private async Task<StaffReposSearchResult> Search(StaffSearchRequest request)
...
```
martin-hornsby: This code works, albeit is a bit smelly. Just wondering if this looks cleaner:

var results = await _staffIlrRepository.SearchForLearnerByCertificateReference(request.SearchQuery);

return new StaffReposSearchResult
{
	DisplayEpao = true,
        PageOfResults = results,
	TotalCount = results.Count()
};

PR #594, opened 2019-02-25
Apr 162 download roatp excel

David-B-Read:
Add link to RoATP landing page to download the RoATP complete register and audit history as an Excel spreadsheet

Has a dependency on the following PRs:

https://github.com/SkillsFundingAgency/das-employer-config/pull/211

https://github.com/SkillsFundingAgency/das-roatp-service/pull/17


src/SFA.DAS.AssessorService.Web.Staff/Controllers/ExcelAwareController.cs
```diff
@@ -0,0 +1,34 @@
+﻿namespace SFA.DAS.AssessorService.Web.Staff.Controllers
+{
+    using Microsoft.AspNetCore.Mvc;
+    using System.Collections.Generic;
+    using System.Data;
+    using System.Linq;
+
+    public class ExcelAwareController : Controller
```
MarkFCain: This is a weird name for a controller.  Is there a reason to call it ExcelAware?

src/SFA.DAS.AssessorService.Web.Staff/Domain/Roles.cs
```diff
@@ -11,15 +11,31 @@ public static class Roles
...
                    || user.IsInRole(AssessmentDeliveryTeam)
                    || user.IsInRole(ProviderRiskAssuranceTeam)
-                   || user.IsInRole(RegisterViewOnlyTeam);
+                   || user.IsInRole(RegisterViewOnlyTeam)
+                   || user.IsInRole(RoatpGatewayTeam);
+        }
+
+        public static bool HasRoatpRoleOnly(this ClaimsPrincipal user)
+        {
+            if (user.IsInRole(RoatpGatewayTeam)
```
MarkFCain: Just a bit worried that this is fragile to new roles being added.  Not sure if there is a solution though....

src/SFA.DAS.AssessorService.Web.Staff/Controllers/ExcelAwareController.cs
David-B-Read: I wanted to reuse the excel data table functionality so made a base class for ReportsController and RoatpController. Happy to go with other suggestions on naming

PR #597, opened 2019-03-01
ON-1556 & ON-1684 & ON-1719 & ON-1758 : Gather apply information inject into assessor

MarkFCain:


martin-hornsby:
I've added additional functionality to inject the organisation standard fi it is an initial application.

Also fixed a few issues injecting the appropriate Contact with the Organisation & Standard.

**This branch will likely need additional work to cater for ASLogin**

martin-hornsby:
I've added functionality to inject the organisation standard when it is not an initial application.

martin-hornsby:
**NOTE:** When we have EPAOrgId within Apply Organisation the following needs changing in `GatherAnswersForOrganisationStandardForApplication` 

```
var assessorOrganisation = (await _assessorApiClient.SearchOrganisations(organisation.Name)).FirstOrDefault();
var organisationId = assessorOrganisation?.Id;
```

martin-hornsby:
ON-1758 added as it was apparent that FHA Details were not being injected into RoEPAO

src/SFA.DAS.AssessorService.Application.Api.UnitTests/Services/AssessorValidationServiceTests.cs
```diff
@@ -123,5 +123,15 @@ public void CheckIfUkprnAlreadyTaken(bool result)
...
+
+        [TestCase(true)]
+        [TestCase(false)]
+        public void IsOrganisationStandardTaken(bool result)
+        {
+            _registerValidationRepository.Setup(r => r.EpaOrganisationStandardExists(It.IsAny<string>(), It.IsAny<int>()))
+                .Returns(Task.FromResult(result));
+            var checkResult = _assessorValidationService.IsOrganisationStandardTaken("EPA0001", 1).Result;
+            Assert.AreEqual(checkResult, result);
+        }
```
davidgouge: How can both of those TestCases pass?

davidgouge: Oh I see, the .Setup has it's return set to the expected.  Not sure this is a very good test, but not sure why.

davidgouge: Would the better test be to make sure that the correct values get passed through to EpaOrganisationStandardExists? 
And check the return from EpaOrganisationStandardExists gets returned from IsOrganisationStandardTaken is the same (which i guess is what this test is doing, but it doesn't need to do both test cases).

src/SFA.DAS.AssessorService.Data/RegisterQueryRepository.cs
```diff
@@ -374,6 +373,40 @@ public async Task<EpaContact> GetContactByContactId(Guid contactId)
...
+
+                var sql =
+                    "select Id, EndPointAssessorOrganisationId, Username,GivenNames, DisplayName, FamilyName, SigninId, SigninType, Email, Status, PhoneNumber " +
+                    " from Contacts where Email = @email";
+                var contact = await connection.QuerySingleOrDefaultAsync<EpaContact>(sql, new { email });
+                return contact;
+            }
+        }
+
+        public async Task<EpaContact> GetContactBySignInId(string signinId)
```
davidgouge: Why is this returning an EpaContact and not a Contact?  And surely we have a GetContactBySignInId somewhere else?  Same for Email as well I guess?

src/SFA.DAS.AssessorService.Web.Staff/Controllers/Apply/ApplicationController.cs
```diff
@@ -258,12 +262,29 @@ public async Task<IActionResult> Return(Guid applicationId, int sequenceId, stri
             var warningMessages = new List<string>();
             if (sequenceId == 2 && returnType == "Approve")
             {
-                var command = await _answerService.GatherAnswersForOrganisationAndContactForApplication(applicationId);
-                var response = await _answerInjectionService.InjectApplyOrganisationAndContactDetailsIntoRegister(command);
-                warningMessages = response.WarningMessages;
-                if (warningMessages.Count == 0 && !response.IsEpaoApproved)
+                var sequneceOne = await _applyApiClient.GetSequence(applicationId, 1);
```
davidgouge: Typo. :)

martin-hornsby: Good spot!

src/SFA.DAS.AssessorService.Data/RegisterQueryRepository.cs
martin-hornsby: Probably as Mark said, that `EPA distinguished itself as being dappery, when there was a lot of EF still in use`
Assuming it's going to be a bit of a job to clean this up into 1 common object again

src/SFA.DAS.AssessorService.Application.Api.UnitTests/Services/AssessorValidationServiceTests.cs
martin-hornsby: It follows the same pattern as the other tests in here (i.e. true / false)

I read these tests as the registerValidationRepository will return the appropraite response and thus assessorValidationService should return that response too (i.e. it bubbles up through the relevant layers in the application).

PR #604, opened 2019-03-11
Apr 233 add organisation validation

David-B-Read:
Add validation and duplicate checks for the add RoATP organisation functionality


src/SFA.DAS.AssessorService.Web.Staff/Controllers/Roatp/AddRoatpOrganisationController.cs
```diff
@@ -138,8 +138,8 @@ private Organisation CreateOrganisationFromModel(AddOrganisationViewModel model)
             var organisation = new Organisation
             {
                 Id = Guid.NewGuid(),
-                LegalName = model.LegalName,
-                TradingName = model.TradingName,
+                LegalName = model.LegalName.ToUpper(),
+                TradingName = model.TradingName?.ToUpper(),
```
MarkFCain: Minor quibble but I'll leave this to you.  We store the legal name in upper case, but Trading name is usually stored in pascal case. Not sure where that convention came from, but it seems to be the pattern?  I'll leave it with you

David-B-Read: It's a requirement in the user story to store both legal name and trading name in upper case

src/SFA.DAS.AssessorService.Web.Staff/Validators/Roatp/AddOrganisationProviderTypeViewModelValidator.cs
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        private ValidationResponse IsValidProviderType(AddOrganisationProviderTypeViewModel viewModel)
+        {
+            var validationResult = new ValidationResponse
+            {
+                Errors = new List<ValidationErrorDetail>()
+            };
+            
+            var isValid = (viewModel.ProviderTypeId >= 1 && viewModel.ProviderTypeId <= 3);
```
MarkFCain: We'll do this some time as tech debt, but I'd like to isolate the rules for validation such as this one into its own service at some point.

src/SFA.DAS.AssessorService.Web.Staff/Validators/Roatp/AddOrganisationValidator.cs
```diff
@@ -2,40 +2,238 @@
...
+
+            fieldValidationErrors = IsValidTradingName(viewModel.TradingName);
+            if (fieldValidationErrors.Any())
+            {
+                validationResponse.Errors.AddRange(fieldValidationErrors);
+            }
+
+            return validationResponse;
+        }
+
```
MarkFCain: See previous comment about isolating these rules into their own service.  Let's create a tech debt story about it soon.

src/SFA.DAS.AssessorService.Web.Staff/Validators/Roatp/AddOrganisationProviderTypeViewModelValidator.cs
David-B-Read: Agreed, I'd also like to put frequently used validation items such as UKPRN, Companies House Number into a Nuget package that can be shared with EPAO and other ESFA projects

PR #606, opened 2019-03-13
Apr 328 create roatp data for fat

MarkFCain:
This is a complement to a roatp story (which will need to go in first)


MarkFCain:
this is no longer needed, as we'll deal with the excel in roatp

PR #615, opened 2019-03-20
ON-1657 - Add standard level everywhere

martin-hornsby:


alandburns:
The Staff changes for this branch need to be re-implemented in das-admin-service

martin-hornsby:
I won't close this yet, however it's looking very likely that this PR should be discarded and for the ticket to be reworked across the various systems

alandburns:
Back to the drawing-board for this one.
Needs UI designs.

PR #623, opened 2019-03-26
Apr 288 update trading name

MarkFCain:
Currently comparing on APR-165, will reset to master once 165 is in master


src/SFA.DAS.AssessorService.Web.Staff/Validators/Roatp/UpdateOrganisationTradinglNameViewModelValidator.cs
```diff
@@ -0,0 +1,26 @@
+﻿namespace SFA.DAS.AssessorService.Web.Staff.Validators.Roatp
```
David-B-Read: Typo in the filename - class name is ok though

src/SFA.DAS.AssessorService.Web.Staff/Views/Roatp/UpdateOrganisationTradingName.cshtml
```diff
@@ -0,0 +1,67 @@
+﻿@*@{
```
David-B-Read: Can this be removed?

PR #631, opened 2019-03-29
Apr 290 update financial track rec

MarkFCain:


src/SFA.DAS.AssessorService.Web.Staff/Controllers/Roatp/UpdateRoatpOrganisationController.cs
```diff
@@ -64,6 +64,52 @@ private UpdateOrganisationLegalNameRequest CreateUpdateLegalNameRequest(UpdateOr
...
-      
+
+        [Route("change-financial-track-record")]
+        public async Task<IActionResult> UpdateOrganisationFinancialTrackRecord()
+        {
+            var searchModel = _sessionService.GetSearchResults();
+
+            var model = new UpdateOrganisationFinancialTrackRecordViewModel
+            {
+                CurrentFinancialTrackRecord = searchModel.SelectedResult.OrganisationData.FinancialTrackRecord,
```
David-B-Read: This is posted back to the controller but doesn't seem to be used?

PR #635, opened 2019-04-02
APR-291 code complete

MarkFCain:


src/SFA.DAS.AssessorService.Web.Staff/Controllers/Roatp/UpdateRoatpOrganisationController.cs
```diff
@@ -64,6 +64,182 @@ private UpdateOrganisationLegalNameRequest CreateUpdateLegalNameRequest(UpdateOr
...
+
+            if (result)
+            {
+                return await RefreshSearchResults();
+            }
+
+            return View("~/Views/Roatp/UpdateOrganisationUkprn.cshtml", model);
+        }
+
+        private UpdateOrganisationUkprnRequest CreateUpdateUkprnRequest(UpdateOrganisationUkprnViewModel model)
```
David-B-Read: Just had a thought - these private methods to map a model to a request could be in a separate class as there's quite a few of them now

PR #654, opened 2019-04-11
On 1750 handle epao users that exist in as login

davidgouge:


src/SFA.DAS.AssessorService.Application.Api/Controllers/OrganisationQueryController.cs
```diff
@@ -92,6 +92,35 @@ public async Task<IActionResult> GetOrganisation(Guid id)
 
             return Ok(organisation);
         }
-        
+
+        [HttpGet("{name}", Name = "GetOrganisationByName")]
```
martin-hornsby: Should be {*name} then you can remove the URL encode/decoding

src/SFA.DAS.AssessorService.Application.Api.Client/Clients/OrganisationsApiClient.cs
```diff
@@ -31,6 +32,16 @@ public async Task<IEnumerable<OrganisationResponse>> GetAll()
             }
         }
 
+        public async Task<OrganisationResponse>  GetOrganisationByName(string name)
+        {
+            using (var request = new HttpRequestMessage(HttpMethod.Get, $"/api/v1/organisations/{WebUtility.UrlEncode(name)}"))
```
martin-hornsby: First time we've had to URL encode something.

Why not use the inbuild MVC stuff? See comment below :)

src/SFA.DAS.AssessorService.Application.Api/Controllers/OrganisationQueryController.cs
```diff
@@ -92,6 +92,35 @@ public async Task<IActionResult> GetOrganisation(Guid id)
...
+        [SwaggerResponse((int)HttpStatusCode.InternalServerError, Type = typeof(ApiResponse))]
+        public async Task<IActionResult> GetOrganisationByName(string name)
+        {
+            var decodedName = WebUtility.UrlDecode(name);
+            _logger.LogInformation($"Received request to retrieve Organisation {decodedName}");
+            
+            var organisation = await _organisationQueryRepository.GetOrganisationByName(decodedName);
+            if(organisation == null)
+            {
+                var ex = new ResourceNotFoundException(name);
```
martin-hornsby: can this be tidied up?

src/SFA.DAS.AssessorService.Web/StartupConfiguration/AuthenticationStartup.cs
```diff
@@ -179,6 +274,22 @@ public static class AuthenticationStartup
                     });
             });
         }
+
+        private async static Task<ContactResponse> CreateANewContact(Contact newContact, IContactsApiClient contactClient, ILogger<Startup> logger, string signInId)
+        {
```
martin-hornsby: Does this need a unit test?

src/SFA.DAS.AssessorService.Web/StartupConfiguration/AuthenticationStartup.cs
```diff
@@ -32,132 +31,228 @@ public static class AuthenticationStartup
...
+                                 //Todo: Need to determine privileges dynamically
+                                 identity.AddClaim(new Claim("http://schemas.portal.com/service",
+                                     Privileges.ManageUsers));
+                             }
+                             
+                             context.Principal.AddIdentity(identity);
+                         }
+
+                     };
+                 });
```
davidgouge: There is *a lot* going on in this Startup class now.  Can this be extracted someplace else to be unit tested?

src/SFA.DAS.AssessorService.Web/StartupConfiguration/AuthenticationStartup.cs
davidgouge: Agree

src/SFA.DAS.AssessorService.Application.Api.Client/Clients/OrganisationsApiClient.cs
davidgouge: Yes, we don't need to be manually encoding/decoding.

arshedmust: Had a situation when some company names where not  being matched since the default action did not encode special characters properly. But using the manual encode/decode did the job.

src/SFA.DAS.AssessorService.Application.Api/Controllers/OrganisationQueryController.cs
arshedmust: Done some tidy up

PR #662, opened 2019-04-12
Apr 390 update provider status

MarkFCain:


src/SFA.DAS.AssessorService.Web.Staff/Infrastructure/RoatpApiClient.cs
```diff
@@ -10,6 +10,7 @@
     using Newtonsoft.Json;
     using SFA.DAS.AssessorService.Api.Types.Models.Roatp;
     using System.Net;
+    using Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal;
```
David-B-Read: Is this used?

PR #667, opened 2019-04-15
Added two new fields First Name and Last Name to staff app create new…

arshedmust:
… contact screen. Removed display name field from screen. This is now generated by combining first name and last name.


martin-hornsby:
Could you correct the varibles in the PR to `GivenNames` and `FamilyName`

This makes it consistent with all the Contact object across our Projects

PR #669, opened 2019-04-15
Apr 291 update organisation ukprn

David-B-Read:


src/SFA.DAS.AssessorService.Web.Staff/Controllers/Roatp/UpdateRoatpOrganisationController.cs
```diff
@@ -269,12 +304,8 @@ public async Task<IActionResult> UpdateProviderType(UpdateOrganisationProviderTy
             var request = Mapper.Map<UpdateOrganisationProviderTypeRequest>(model);
             var result = await _apiClient.UpdateOrganisationProviderType(request);
 
-            if (result)
-            {
-                return await RefreshSearchResults();
-            }
-
```
MarkFCain: put it back!

PR #672, opened 2019-04-15
Public beta assessor service

alandburns:
Hot Fix applied


alandburns:
Bringing latest Public Beta Branch changes back to master

PR #676, opened 2019-04-16
On 1777 correct dashboard pipeline counts

martin-hornsby:


src/SFA.DAS.AssessorService.Application/Interfaces/IStandardRepository.cs
```diff
@@ -14,7 +14,6 @@ public interface IStandardRepository
         Task<StandardCollation> GetStandardCollationByReferenceNumber(string referenceNumber);
         Task<DateTime?> GetDateOfLastStandardCollation();
         Task<int> GetEpaoStandardsCount(string endPointAssessorOrganisationId);
```
martin-hornsby: GetEpaoStandardsCount is used elsewhere. Left it in the Interface so that we can unit test it

src/SFA.DAS.AssessorService.Application.Api.UnitTests/Controllers/Organisations/Query/OrganisationQueryBase.cs
```diff
@@ -47,10 +47,7 @@ protected  void Setup()
             {
                 cfg.CreateMap<Organisation, OrganisationResponse>();
             });
-            EpaoStandardsCountResponse response = new EpaoStandardsCountResponse(1);
-         
-            Mediator.Setup(q => q.Send(Moq.It.IsAny<GetEpaoStandardsCountRequest>(), new CancellationToken()))
```
martin-hornsby: The unit tests do not depend on this so removed it.

PR #677, opened 2019-04-17
On 1769 error when adding contact staff app

arshedmust:
I have reverted this branch from master and rebased it with public beta


alandburns:
Not needed as all changes in ON-1780

PR #684, opened 2019-04-17
On 1780 edit contact page in staff app

arshedmust:


alandburns:
Although this change did not create the issue, we should not be using GET to pass personal details.
It is a security risk.
It should be a POST.

martin-hornsby:
> The ValidateCreateContact should be using a POST not a GET. personal details should not be passed on the Query string and it is an action (validate) not a simple GET (read).  Same for ValidateUpdateContact.
>         using (var request = new HttpRequestMessage(HttpMethod.Get,
>             $"/api/ao/assessment-organisations/contacts/validate-new?firstName={newFirstName}&lastName={newLastName}&organisationId={organisationId}&email={newEmail}&phone={phone}"))

I've updated those end points to reflect the comment Alan made

PR #686, opened 2019-04-18
APR-294 code done

MarkFCain:


src/SFA.DAS.AssessorService.Web.Staff/Controllers/Roatp/UpdateRoatpOrganisationController.cs
```diff
@@ -158,6 +176,29 @@ public async Task<IActionResult> UpdateStatus(UpdateOrganisationStatusViewModel
             return View("~/Views/Roatp/UpdateOrganisationStatus.cshtml", model);
         }
 
+        [HttpPost]
+        public async Task<IActionResult> UpdateType(UpdateOrganisationTypeViewModel model)
+        {
+            model.OrganisationTypes = _apiClient.GetOrganisationTypes(model.ProviderTypeId).Result.Where(x=>x.Id!=0).OrderBy(x => x.Id);
```
David-B-Read: Thinking it would be a good idea to have the organisation type id 0 (Unassigned) as a constant so it's easy to see that we are excluding that particular type

src/SFA.DAS.AssessorService.Web.Staff/Views/Roatp/UpdateOrganisationType.cshtml
```diff
@@ -0,0 +1,83 @@
...
+        </h2>
+        <ul class="error-summary-list">
+            <li><a sfa-validation-for="OrganisationTypeId"></a></li>
+        </ul>
+    </div>
+}
+
+<div class="grid-row">
+    <div class="column-two-thirds">
+        <h1 class="heading-xlarge roatp-heading">Change Type for @Html.Raw(Model.LegalName)</h1>
```
David-B-Read: As part of the fix for APR-420 I've annotated the legal name in the headings for the update pages so that long names wrap correctly, like the example below: 

A quick win to add this then that potential bug gets fixed

`<h1 class="heading-xlarge roatp-heading">Change financial track record for <span class="roatp-update-organisation-long-text">@Html.Raw(Model.LegalName)</span></h1>`

MarkFCain: will do

src/SFA.DAS.AssessorService.Web.Staff/Controllers/Roatp/UpdateRoatpOrganisationController.cs
MarkFCain: yeah, there's 2 places I've used it, and I've added it as a constant to the OrganisationType class and using that instead.  Much better than magic numbers

PR #688, opened 2019-04-18
ON-1759 - Show external API link on dashboard if their org has an UKPRN

arshedmust:
…h enabling user as a primary contact


src/SFA.DAS.AssessorService.Web/StartupConfiguration/AuthenticationStartup.cs
```diff
@@ -218,33 +230,32 @@ public static class AuthenticationStartup
...
                                          identity.AddClaim(new Claim("http://schemas.portal.com/ukprn",
-                                            organisation?.Ukprn == null ? "" : organisation?.Ukprn.ToString()));
+                                             organisation?.Ukprn == null ? "" : organisation?.Ukprn.ToString()));
 
-                                         var orgName = organisation.OrganisationData?.LegalName ??
-                                                       organisation.OrganisationData?.TradingName ??
-                                                       organisation.Name;
+                                         var orgName = organisation?.OrganisationData?.LegalName ??
+                                                       organisation?.OrganisationData?.TradingName ??
+                                                       organisation?.Name;
```
davidgouge: From @alandburns 

> How is it possible to have No organisation?
> And if there is no value in the orgName claim this will cause an Unhandled Exception.

src/SFA.DAS.AssessorService.Application.Api/Controllers/ContactQueryController.cs
```diff
@@ -79,7 +79,7 @@ public async Task<IActionResult> GetAllContactsForOrganisation(string endPointAs
             return Ok(contacts);
         }
 
-        [HttpGet("user/{userName}", Name = "SearchContactByUserName")]
+        [HttpGet("username/{userName}", Name = "SearchContactByUserName")]
```
martin-hornsby: This is required changed as it conflicted with [HttpGet("user/{id}", Name = "GetContactById")]

src/SFA.DAS.AssessorService.Web/StartupConfiguration/AuthenticationStartup.cs
martin-hornsby: fixed this with a null check futher up

martin-hornsby:
I think this PR will need to be revisited as without it I am unable to test External API without hacking permissions in my debugger and typing in URLs directly

alandburns:
HAs been redeveloped in new User Management ON-1208

PR #690, opened 2019-04-24
Apr 438 pen test fix html injection

MarkFCain:


src/SFA.DAS.AssessorService.Web.Staff/Controllers/Roatp/AddRoatpOrganisationController.cs
```diff
@@ -90,7 +91,8 @@ public async Task<IActionResult> AddOrganisationPreview(AddOrganisationViewModel
         {
             model.OrganisationTypes = await _apiClient.GetOrganisationTypes(model.ProviderTypeId);
             model.ProviderTypes = await _apiClient.GetProviderTypes();
-
+            model.LegalName = HtmlTagRemover.StripOutTags(model?.LegalName);
```
David-B-Read: One for the tech debt story, but should the stripping of the tags be done at the same time that the legal name is converted to upper case?

PR #707, opened 2019-05-02
ON-1737 Update order and display of user details consistently with apply

gdbarnes:


src/SFA.DAS.AssessorService.Web/Controllers/AccountController.cs
```diff
@@ -51,11 +55,14 @@ public IActionResult SignIn()
         public async Task<IActionResult> PostSignIn()
         { 
             var loginResult = await _loginOrchestrator.Login();
+            var orgName = _contextAccessor.HttpContext.User.FindFirst("http://schemas.portal.com/orgname")?.Value;
```
martin-hornsby: Is it always in the claim? If not then I'd be tempted to use the Orgname property from loginResult

src/SFA.DAS.AssessorService.Application/Handlers/Login/LoginHandler.cs
```diff
@@ -118,8 +116,6 @@ public async Task<LoginResponse> Handle(LoginRequest request, CancellationToken
                 }
             }
 
-            _logger.LogInformation(LoggingConstants.SignInSuccessful);
```
martin-hornsby: Assuming this is now redundant, although isn't it useful knowing it was successful in the logs?

src/SFA.DAS.AssessorService.Web/Controllers/AccountController.cs
ChrisWoodcock: Prior to this code being activated the claim is populated with the organisation name which is passed to apply - subsequently assessor didn't use the value it had passed to apply

src/SFA.DAS.AssessorService.Application/Handlers/Login/LoginHandler.cs
ChrisWoodcock: It's actually repeated above - you can't see it in the diff view - but the result would have been Successful Successful in the log - so I removed the duplicate.

PR #718, opened 2019-05-08
ON-1839 - Check company is actively trading

martin-hornsby:


src/SFA.DAS.AssessorService.Application.Api.Client/Clients/ApiClientBase.cs
```diff
@@ -84,7 +84,7 @@ private static HttpRequestException CreateRequestException(HttpRequestMessage fa
                     failedResponse.Content.ReadAsStringAsync().Result));
         }
 
-        protected async Task<T> RequestAndDeserialiseAsync<T>(HttpRequestMessage request, string message = null) where T : class
```
martin-hornsby: Structs are valid JSON too! We have instances where an API call returns a struct  (i.e. true).

I've removed the constraint as otherwise I'd be creating an identical version of this method for structs.

src/SFA.DAS.AssessorService.Web/Controllers/OrganisationSearchController.cs
```diff
@@ -112,6 +112,16 @@ public async Task<IActionResult> Confirm(OrganisationSearchViewModel viewModel)
 
             if (organisationSearchResult != null)
             {
+                if (organisationSearchResult.CompanyNumber != null)
+                {
+                    var isActivelyTrading = await _organisationsApplyApiClient.IsCompanyActivelyTrading(organisationSearchResult.CompanyNumber);
+
+                    if (!isActivelyTrading)
+                    {
+                        return View("~/Views/OrganisationSearch/CompanyNotActive.cshtml", viewModel);
```
davidgouge: Shouldn't _really_ be returning a View from a POST (should be returning a redirect to a GET).  But if the rest of this method is doing the same... 🤷‍♂

PR #722, opened 2019-05-09
Apr 292 update company number

MarkFCain:


src/SFA.DAS.AssessorService.Web.Staff/Controllers/Roatp/UpdateRoatpOrganisationController.cs
```diff
@@ -142,6 +142,42 @@ public async Task<IActionResult> UpdateOrganisationType()
             return View("~/Views/Roatp/UpdateOrganisationType.cshtml", model);
         }
 
+
+        [Route("change-companies-number")]
```
David-B-Read: URL should be change-company-number, as per content spreadsheet
https://docs.google.com/spreadsheets/d/1EtRem3hbvkPos-AFNbf9p_QQzJS6WQWLwkKO2POif2A/edit#gid=0

MarkFCain: yep.  Now fixed.

PR #730, opened 2019-05-14
On 1865 replace fat calls with local db

davidgouge:
Changes to remove all calls to the FAT api except those related to gathering standards to update our StandardCollations table.  

Removed the old EpaoImporter function and related Handlers.

Refactored StandardService and removed usage of FAT Api types like Standard and StandardSummary.


src/SFA.DAS.AssessorService.Application.Api/Controllers/StandardServiceController.cs
```diff
@@ -40,7 +33,7 @@ public StandardServiceController(ILogger<StandardServiceController> logger, ISta
         public async Task<IActionResult> StandardServiceGetAllStandardSummaries()
         {
             _logger.LogInformation($@"Get all StandardSummaries from Standard Service");
```
martin-hornsby: Do we need this as it used to return a list of StandardSummary
Think the naming of this method going forward will get confusing

src/SFA.DAS.AssessorService.Application.Api/Services/StandardService.cs
```diff
@@ -30,78 +30,33 @@ public StandardService(CacheService cacheService, IAssessmentOrgsApiClient asses
             _standardRepository = standardRepository;
         }
 
-        public async Task<IEnumerable<StandardSummary>> GetAllStandardsV2()
+        public async Task<IEnumerable<StandardCollation>> GetAllStandardsV2()
```
martin-hornsby: Do we need it to say V2 still?

src/SFA.DAS.AssessorService.Application/Handlers/Search/SearchResultExtensions.cs
```diff
@@ -17,17 +17,17 @@ static class SearchResultExtensions
     {
         public static List<SearchResult> PopulateStandards(this List<SearchResult> searchResults, IStandardService standardService, ILogger<SearchHandler> logger)
         {
-            var allStandards = standardService.GetAllStandards().Result;
+            var allStandards = standardService.GetAllStandardsV2().Result;
 
             foreach (var searchResult in searchResults)
             {
-                var standard = allStandards.SingleOrDefault(s => s.StandardId == searchResult.StdCode.ToString());
+                var standard = allStandards.SingleOrDefault(s => s.ReferenceNumber == searchResult.StdCode.ToString());
```
martin-hornsby: This looks incorrect. StdCode is usually StandardId

src/SFA.DAS.AssessorService.Application/Interfaces/IStandardService.cs
```diff
@@ -7,12 +7,12 @@ namespace SFA.DAS.AssessorService.Application.Interfaces
 {
     public interface IStandardService
     {
-        Task<IEnumerable<StandardSummary>> GetAllStandardsV2();
-        Task<IEnumerable<Standard>> GetAllStandards();
-        Task<IEnumerable<StandardCollation>> GatherAllStandardDetails();
-        Task<Standard> GetStandard(int standardId);
-        Task<Standard> GetStandard(string referenceNumber);
-        Task<IEnumerable<StandardSummary>> GetAllStandardSummaries();
+        Task<IEnumerable<StandardCollation>> GetAllStandardsV2();
```
martin-hornsby: As mentioned earlier do we really want to call this V2?

src/SFA.DAS.AssessorService.Application.Api/Services/StandardService.cs
davidgouge: Ahh! Good one, was going to rename that. 👍

src/SFA.DAS.AssessorService.Application/Interfaces/IStandardService.cs
davidgouge: ☝️

src/SFA.DAS.AssessorService.Application.Api/Controllers/StandardServiceController.cs
davidgouge: Yes, good point.  Will rename.

src/SFA.DAS.AssessorService.Application/Handlers/Search/SearchResultExtensions.cs
davidgouge: Good spot, thanks. 😁

src/SFA.DAS.AssessorService.Application.Api/Controllers/StandardServiceController.cs
```diff
@@ -37,18 +30,18 @@ public StandardServiceController(ILogger<StandardServiceController> logger, ISta
         [SwaggerResponse((int)HttpStatusCode.OK, Type = typeof(IEnumerable<StandardSummary>))]
```
martin-hornsby: Is this the correct return type?

PR #732, opened 2019-05-15
ON-1848 Changes to Manage Users page

davidgouge:
Change user name to link to a (currently non-existant) User details page.
Added button to (currently non-existant) Invite User page

### DIFF
--- a/src/SFA.DAS.AssessorService.Web/Controllers/ManageUsers/InviteUserController.cs
+++ b/src/SFA.DAS.AssessorService.Web/Controllers/ManageUsers/InviteUserController.cs
@@ -1,0 +1,23 @@
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using SFA.DAS.AssessorService.Web.Infrastructure;
+
+namespace SFA.DAS.AssessorService.Web.Controllers.ManageUsers
+{
+    [Authorize]
+    [CheckSession]
+    public class InviteUserController : Controller
+    {
+        public InviteUserController()
+        {
+            
+        }
+
+        [HttpGet("/ManageUsers/Invite")]
+        public async Task<IActionResult> Invite()
+        {
+            return View();
+        }
+    }
+}
--- a/src/SFA.DAS.AssessorService.Web/Views/ManageUsers/Index.cshtml
+++ b/src/SFA.DAS.AssessorService.Web/Views/ManageUsers/Index.cshtml
@@ -70,7 +70,7 @@
                     </div>
                 </div>
             }
-            <a asp-controller="InviteUser" asp-action="Invite" class="button">Invite new user</a>
+            <a asp-controller="InviteUser" asp-action="Invite" class="button govuk-button">Invite new user</a>
         </div>
     </div>
 </div>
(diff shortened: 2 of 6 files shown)


src/SFA.DAS.AssessorService.Web/Controllers/ManageUsers/InviteUserController.cs
```diff
@@ -0,0 +1,23 @@
...
+using SFA.DAS.AssessorService.Web.Infrastructure;
+
+namespace SFA.DAS.AssessorService.Web.Controllers.ManageUsers
+{
+    [Authorize]
+    [CheckSession]
+    public class InviteUserController : Controller
+    {
+        public InviteUserController()
+        {
```
martin-hornsby: I assume there will be some sort of implementation in here later 😉

src/SFA.DAS.AssessorService.Web/Views/ManageUsers/Index.cshtml
```diff
@@ -31,29 +31,30 @@
                         </tr>
```
martin-hornsby: Whilst not part of this PR, we're exposing and API Type directly to the view. This should be a View Model

src/SFA.DAS.AssessorService.Web/Controllers/ManageUsers/InviteUserController.cs
davidgouge: Yes, as I mentioned, the implementation is in a different ticket.  This is just for the main page.

src/SFA.DAS.AssessorService.Web/Views/ManageUsers/Index.cshtml
davidgouge: On the whole, I agree, but in this case the VM and the type would be identical and it would be a map between the two for no reason.  If that part of the view becomes more complicated and view specific, then yeah, VM all the way.

src/SFA.DAS.AssessorService.Web/Views/ManageUsers/Index.cshtml
```diff
@@ -31,29 +31,30 @@
...
                     <tbody>
-                        @foreach (var item in Model)
+                        @foreach (var contact in Model)
                         {
                             <tr>
-                                <td data-label="@Localizer["User details"]" class="nowrap" >
-                                    @Html.DisplayFor(modelItem => item.Contact.DisplayName)<br />
-                                    <span>@Html.DisplayFor(modelItem => item.Contact.Email)</span>
+                                <td data-label="@Localizer["User details"]" class="nowrap">
+                                    <a asp-controller="ManageUsers" asp-action="User" asp-route-userid="@contact.Contact.Id">@Html.DisplayFor(modelItem => contact.Contact.DisplayName)</a>
```
gdbarnes: This will show the new link correctly when the css upgrade is merged:
```suggestion
                                    <a class="govuk-link" asp-controller="ManageUsers" asp-action="User" asp-route-userid="@contact.Contact.Id">@Html.DisplayFor(modelItem => contact.Contact.DisplayName)</a>
```

src/SFA.DAS.AssessorService.Web/Views/ManageUsers/Index.cshtml
```diff
@@ -69,6 +70,7 @@
                     </div>
                 </div>
             }
+            <a asp-controller="InviteUser" asp-action="Invite" class="button">Invite new user</a>
```
gdbarnes: This will show the button correctly when the css upgrade is merged:
```suggestion
            <a asp-controller="InviteUser" asp-action="Invite" class="button govuk-button">Invite new user</a>
```

PR #736, opened 2019-05-16
ON-1874 Standards shown on the standards tab in the dashboard are effective in future or today

ChrisJWoodcock:


martin-hornsby:
Was just thinking, could we make this query tidier (and a tad more performant) by using NULL COALESCE:

```
and ISNULL(os.effectiveTo, SYSDATETIME()) > GETDATE())
and ISNULL(JSON_VALUE(StandardData,'$.EffectiveTo'), SYSDATETIME()) > GETDATE())
```

ChrisJWoodcock:
> Was just thinking, could we make this query tidier by using ISNULL function or COALESCE:
> 
> ```
> AND ISNULL(os.effectiveTo, SYSDATETIME()) > GETDATE())
> AND ISNULL(JSON_VALUE(StandardData,'$.EffectiveTo'), SYSDATETIME()) > GETDATE())
> ```

That is a good idea, however it has been done the long way because that is how it is done in lots of other places; it would be better done everywhere as a function and if it was done even as a function then ISNULL(... , @MAX_DATE) > GETDATE() is better because using SYSDATETIME() as 3 nanoseconds > GETDATE() due to a rounding error isn't clear what is being checked and may change in the future.

PR #739, opened 2019-05-16
On 1208 user management

davidgouge:


davidgouge:
Closed as no longer required

PR #747, opened 2019-05-22
ON-1525 Updated SQL for search for ILRs using uln as per alan recommendation

arshedmust:


arshedmust:
Removed the order statement

PR #748, opened 2019-05-22
Apr 471 ukrlp service is unavailable

MarkFCain:


src/SFA.DAS.AssessorService.Web.Staff/Controllers/Roatp/AddRoatpOrganisationController.cs
```diff
@@ -53,12 +54,22 @@ public async Task<IActionResult> UkprnPreview(AddOrganisationViaUkprnViewModel m
...
             }
-            var details = await _ukrlpClient.Get(model.UKPRN);
 
+            UkrlpProviderDetails details;
+
+            try
+            {
+                details = await _ukrlpClient.Get(model.UKPRN);
+            }
+            catch (Exception)
```
David-B-Read: Can this be a more specific exception handler?

src/SFA.DAS.AssessorService.Web.Staff/Views/Roatp/UkprnIsUnavailable.cshtml
```diff
@@ -0,0 +1,172 @@
+﻿@model SFA.DAS.AssessorService.Web.Staff.ViewModels.Roatp.UkrlpNotFoundViewModel
+
+@{ ViewBag.Title = "UKRLP Not Found"; }
```
David-B-Read: Does this title need updating?

PR #750, opened 2019-05-23
On 1862 remove user

davidgouge:


src/SFA.DAS.AssessorService.Application.UnitTests/Handlers/Contacts/WhenRetrieveContactsWithPrivileges.cs
```diff
@@ -19,48 +19,49 @@ public class WhenRetrieveContactsWithPrivileges
     {
         private List<ContactsWithPrivilegesResponse> _result;
```
martin-hornsby: Is this unit tested needed anymore? Should it be replaced with something else? ;)

PR #753, opened 2019-05-24
Ukrlp branch new flow

MarkFCain:
This is a subbranch based on the first branch APR-458 that will be the basis for a new flow in the roatp entry process.  It will run parallel to master, and is for the ROATP team


MarkFCain:
This is being merged into admin-services, and no longer needed in assessor-service

PR #756, opened 2019-05-24
ON-1568 : External API Data Sync for DEMO playground

martin-hornsby:


martin-hornsby:
Whilst doing required changes to the Domain / DB Context I found a couple of minor issues that should be addressed as Tech Debt. As some of these would cause cascading namespace changes throughout the codebase, I have left things as is due to the fact that I didn't want to make this Pull Request become any bigger that what it already is.

martin-hornsby:
This has been closed. A new branch will be opened in its place

PR #757, opened 2019-05-24
ON-1639 - Update title & link to Completed Assessements

martin-hornsby:


src/SFA.DAS.AssessorService.Web.Staff/Views/Apply/Financial/Graded.cshtml
```diff
@@ -15,7 +15,7 @@
                 <span class="bold">@Model.QnAData.FinancialApplicationGrade.SelectedGrade</span>.
             </p>
         </div>
-        <p>You can view previous evaluations, including this one in the <a asp-action="PreviousApplications" asp-controller="Financial">previous assessments</a> area.</p>
+        <p>You can view previous evaluations, including this one in the <a asp-action="ClosedApplications" asp-controller="Financial">completed assessments</a> area.</p>
```
ChrisJWoodcock: Why is the action method not called CompletedApplications?

martin-hornsby: This has been changed

PR #759, opened 2019-05-28
ON-1637 New search results page layout

arshedmust:


src/SFA.DAS.AssessorService.Web/Views/OrganisationSearch/Results.cshtml
```diff
@@ -75,28 +74,21 @@
                                         <input asp-for="Postcode" value="@result.Address?.Postcode" type="hidden" />
 
                                         <p class="govuk-body govuk-!-margin-top-4">
-                                            <button type="submit" class="govuk-button govuk-button--looks-like-link">@result.Name</button>
+                                            <button type="submit" class="govuk-button govuk-button__looks-like-link--No-Decoration">@result.Name</button>
```
gdbarnes: Why do we need to style this with no underline?

gdbarnes: Actually looking at the ticket I can see you've just followed the design. Maybe make the class lowercase if you can be bothered!

PR #764, opened 2019-05-30
On 1748 new user request access

arshedmust:


src/SFA.DAS.AssessorService.Web/Views/OrganisationSearch/Results.cshtml
```diff
@@ -8,7 +8,7 @@
     bool showErrors = TempData["ShowErrors"] as bool? ?? false;
 }
 
-<a asp-action="Index" asp-controller="OrganisationSearch" class="govuk-back-link">Search again</a>
+@*<a asp-action="Index" asp-controller="OrganisationSearch" class="govuk-back-link">Search again</a>*@
```
ChrisJWoodcock: Did you mean to comment this out? - if it won't be brought back soon it should be removed.

arshedmust: Yes I meant to comment this out. Will see when testers test it if this is still needed. According to the story it has been removed. But user cannot now go back to the original page.

ChrisJWoodcock: You need to query that with a BA and get the AC's to clarify and then remove it or put it back in

alandburns: Let's get this to the testers.

alandburns:
Has been merged to ON-1637

PR #772, opened 2019-06-03
ON-1173 secondary app submissions v2

martin-hornsby:


src/SFA.DAS.AssessorService.Web.Staff/Services/AnswerService.cs
```diff
@@ -22,56 +20,49 @@ public AnswerService(IApplyApiClient applyApiClient, IApiClient assessorApiClien
...
+          
             var command = new CreateOrganisationContactCommand
-            (organisationName,
-                organisationType,
-                organisationUkprn?.ToString(),
-                organisationReferenceType,
-                isEpaoApproved,
+            (   organisation.Name,
+                organisation.OrganisationType,
+                organisation.OrganisationUkprn.ToString(),
```
ChrisJWoodcock: Ukprn could potentially be null

ChrisJWoodcock: Ok

PR #780, opened 2019-06-05
On 1859 invite user to organisation

davidgouge:
Changes to allow a User to Invite another new user to the Organisation. 

Has equivalent changes in both Apply and Login.

There's also some refactoring / renaming in here, so apologies for that.


src/SFA.DAS.AssessorService.Application.Api/Controllers/OrganisationQueryController.cs
```diff
@@ -112,5 +112,17 @@ public async Task<IActionResult> GetOrganisationByName(string name)
             
             return Ok(Mapper.Map<OrganisationResponse>(organisation));
         }
+        
+        [HttpGet("forContact/{userId}")]
+        public async Task<IActionResult> GetOrganisationForContact(Guid userId)
```
martin-hornsby: The routing for this looks weird for a RESTFUL call. Would this be better in the ContactQueryController?

So the routing would be something like: _/contact/{userId}/organisation_

src/SFA.DAS.AssessorService.Application.Api.Client/Clients/OrganisationsApiClient.cs
```diff
@@ -52,6 +62,14 @@ public async Task<OrganisationResponse> Get(string ukprn)
             }
         }
 
+        public async Task<Organisation> Get(Guid organisationId)
+        {
+            using (var request = new HttpRequestMessage(HttpMethod.Get, $"/api/v1/organisations/organisation/{organisationId}"))
```
martin-hornsby: This one looks a bit strange too. Could we not use: _/api/v1/organisations/{organisationId}_

davidgouge: This is an existing endpoint.

src/SFA.DAS.AssessorService.Application.Api/Controllers/OrganisationQueryController.cs
davidgouge: If I was after an organisation, I would expect to find a method to get it in the Organisation controller.  And I think we left RESTful in the dust lonnnnng ago. 😆

martin-hornsby: I would have said it's a Contact's organisation as they own the FK relationship. Anyhow it's works so 👍

src/SFA.DAS.AssessorService.Application.Api.Client/Clients/OrganisationsApiClient.cs
martin-hornsby: Fair enough. Although we need to start addressing these badly named end points at some point 💩

src/SFA.DAS.AssessorService.Application/Handlers/UserManagement/InviteContactToOrganisationHandler.cs
```diff
@@ -0,0 +1,87 @@
...
+                return new InviteContactToOrganisationResponse()
+                {
+                    Success = false,
+                    ErrorMessage = existingContact.OrganisationId == request.OrganisationId
+                        ? "This email address is already registered against your organisation. You must use a unique email address."
+                        : "This email address is already registered against another organisation. You must use a unique email address."
+                };
+            }
+
+            var newContact = await CreateNewContact(request);
```
ChrisJWoodcock: This could be more defensive - there is a lot that could fail here giving nulls and then null references afterwards

src/SFA.DAS.AssessorService.Data/ContactQueryRepository.cs
```diff
@@ -68,12 +67,16 @@ public async Task<Contact> GetContact(string userName)
...
-                .Include(organisation => organisation.Contacts)
-                .Where(q => q.Status != OrganisationStatus.Deleted)
-                .SelectMany(q => q.Contacts)
-                .Where(q => q.Email == email)
-                .FirstOrDefaultAsync();
+            var contact = await _assessorDbContext.Contacts
+                .Include(c => c.Organisation)
+                .FirstOrDefaultAsync(c => c.Email.Equals(email, StringComparison.InvariantCultureIgnoreCase) && c.Organisation.Status != OrganisationStatus.Deleted);
+            
+//            var contact = await _assessorDbContext.Organisations
```
ChrisJWoodcock: Commented out code here

PR #782, opened 2019-06-06
ON-1650 - Show Privately Funded in Certificate Number search

martin-hornsby:


alandburns:
This branch needs to be re-implemented in das-admin-service

martin-hornsby:
Does the new admin service have it's own database repositories?

alandburns:
Needs to be implemented in  https://github.com/SkillsFundingAgency/das-admin-service

alandburns:
OK. I now understand that this is not Staff UI, but at the API in Assessor (for Staff)

PR #784, opened 2019-06-07
On 1857 application of permissions

davidgouge:


src/SFA.DAS.AssessorService.Application/Handlers/UserManagement/RequestForPrivilegeHandler.cs
```diff
@@ -0,0 +1,71 @@
...
+        {
+            _mediator = mediator;
+            _contactQueryRepository = contactQueryRepository;
+            _organisationQueryRepository = organisationQueryRepository;
+            _config = config;
+            _contactRepository = contactRepository;
+        }
+        
+        public async Task Handle(RequestForPrivilegeRequest message, CancellationToken cancellationToken)
+        {
```
martin-hornsby: Do you need to worry if anything returns null (i.e. it cannot find the contact or the privilege for example) ?

src/SFA.DAS.AssessorService.Database/Tables/Privileges.sql
```diff
@@ -1,7 +1,8 @@
 ﻿CREATE TABLE [dbo].[Privileges](
 	[Id] [uniqueidentifier] NOT NULL,
```
martin-hornsby: Could we make Id have a Default of NEWID() ?

src/SFA.DAS.AssessorService.Application/Handlers/UserManagement/RequestForPrivilegeHandler.cs
davidgouge: Not in this case as we are already sure the contact and privilege exist as the Contact is the one trying to access the page (so must exist) and the privilege is what's stopped the contact accessing the page (so must exist). 😁

src/SFA.DAS.AssessorService.Database/Tables/Privileges.sql
davidgouge: Possibly in the future, but for now these Privileges aren't created anywhere in code and are just 'hard coded' into the db. 👍

PR #785, opened 2019-06-07
ON-1940 Organisations exempt from financial's journey not working as it should

arshedmust:


arshedmust:
Updated so that the FinancialExempt is defaulted to false if Organisation Type not set or is not one of the Exempt Types.

PR #787, opened 2019-06-11
Apr 494 application determined date in add provider journey

MarkFCain:


src/SFA.DAS.AssessorService.Web.Staff/Controllers/Roatp/AddRoatpOrganisationController.cs
```diff
@@ -376,6 +339,129 @@ public async Task<IActionResult> Back(string action, Guid organisationId)
...
+            return !ModelState.IsValid
+                ? ModelState.SelectMany(k => k.Value.Errors.Select(e => new ValidationErrorDetail()
+                {
+                    ErrorMessage = e.ErrorMessage,
+                    Field = k.Key
+                })).ToList()
+                : null;
+        }
+
+        private static AddOrganisationTypeViewModel MapOrganisationVmToOrganisationTypeVm(AddOrganisationViewModel addOrganisationModel)
```
David-B-Read: Can we use AutoMapper here?

src/SFA.DAS.AssessorService.Web.Staff/Controllers/Roatp/AddRoatpOrganisationController.cs
```diff
@@ -376,6 +339,129 @@ public async Task<IActionResult> Back(string action, Guid organisationId)
...
+                TradingName = addOrganisationModel.TradingName,
+                UKPRN = addOrganisationModel.UKPRN,
+                Day = determinedDay,
+                Month = determinedMonth,
+                Year = determinedYear
+            };
+        }
+
+
+        private static AddOrganisationProviderTypeViewModel MapOrganisationVmToProviderTypeVm(AddOrganisationViewModel addOrganisationModel)
```
David-B-Read: Can we use AutoMapper here?

src/SFA.DAS.AssessorService.Web.Staff/Controllers/Roatp/AddRoatpOrganisationController.cs
MarkFCain: oooh good call. Will do

PR #799, opened 2019-06-19
ON-1958 - Added several fixes to stop user seeing data they shouldn't

martin-hornsby:


martin-hornsby:
Have checked it, I took out the filter for AccessDenied. The others are based upon having Contact & Organisation information associated with the user

PR #800, opened 2019-06-19
Apr 495 application determined date updated for existing providers

MarkFCain:


src/SFA.DAS.AssessorService.Web.Staff.Tests/Controllers/Roatp/AddRoatpOrganisationControllerTests.cs
```diff
@@ -121,6 +121,16 @@ public void Add_provider_type_model_without_legal_name_redirect_to_organisation_
             _client.VerifyAll();
         }
 
+        [Test]
+        public void Add_model_with_invalid_application_determined_date_redirect_to_xxx()
```
David-B-Read: Is this test name accurate, the expectation is for a view result, not a redirect?

src/SFA.DAS.AssessorService.Web.Staff/Validators/Roatp/UpdateApplicationDeterminedDateViewModelValidator.cs
```diff
@@ -0,0 +1,82 @@
...
+
+            var validationResult = new ValidationResponse
+            {
+                Errors = new List<ValidationErrorDetail>()
+            };
+
+            if (viewModel.Day == null && viewModel.Month == null && viewModel.Year == null)
+            {
+                validationResult.Errors.Add(new ValidationErrorDetail(dateControlName, RoatpOrganisationValidation.ApplicationDeterminedDateNoFieldsEntered));
+                validationResult.Errors.Add(new ValidationErrorDetail("Day", "day error"));
```
David-B-Read: Should these error messages be in a resource file?

MarkFCain: The words for these are actually not used (they defer to the overacrching control message) but I'll think I'll move them to a resource anyway as a bit of future proofing

PR #803, opened 2019-06-20
On 1905 enable inprocess hosting

ChrisJWoodcock:


alandburns:
We need to wait for .netcore3.0 to safely enable InProcess hosting, as several bugs in .netcore2.2 have already been fixed in 3.0

PR #807, opened 2019-06-21
ON-1987 Allow Staff users to view feedback on Application without a Financial…

alandburns:
… grading


alandburns:
Data has been fixed.
I assume this was only allowed by a historical bug, and this should not be necessary.

PR #815, opened 2019-06-28
On 1999 print function not updating certificate logs

arshedmust:


src/SFA.DAS.AssessorService.EpaoImporter/PrintProcessCommand.cs
```diff
@@ -135,49 +135,65 @@ private async Task UploadCertificateDetailsToPinter()
                 var batchLogResponse = await _assessorServiceApi.GetCurrentBatchLog();
 
                 var batchNumber = batchLogResponse.BatchNumber + 1;
-                var certificates = (await _assessorServiceApi.GetCertificatesToBePrinted()).ToList()
-                    .Sanitise(_aggregateLogger);
+                var certificates = (await _assessorServiceApi.GetCertificatesToBePrinted()).ToList();
```
martin-hornsby: We're not Sanitising bad certificates any more?

src/SFA.DAS.AssessorService.EpaoImporter/PrintProcessCommand.cs
```diff
@@ -135,49 +135,65 @@ private async Task UploadCertificateDetailsToPinter()
...
 
-                    var batchLogRequest = new CreateBatchLogRequest
-                    {
-                        BatchNumber = batchNumber,
-                        FileUploadStartTime = DateTime.UtcNow,
-                        Period = DateTime.UtcNow.UtcToTimeZoneTime().ToString("MMyy"),
-                        BatchCreated = DateTime.UtcNow,
-                        CertificatesFileName = certificateFileName
-                    };
+                    await _assessorServiceApi.UpdateBatchNumberInCertificates(batchNumber, certificates);
```
martin-hornsby: This looks like it will update the batch number in Certificates which are not deemed as 'sanitised'

src/SFA.DAS.AssessorService.Domain/Consts/CertificateStatus.cs
```diff
@@ -15,5 +15,6 @@ public class CertificateStatus
         public const string Deleted = "Deleted";
         public const string Reprint = "Reprint";
         public const string Error = "Error";
+        public const string Queued = "Queued";
```
martin-hornsby: This new status needs to be reflected in other parts of the system to ensure Certificates are in the correct status so they can be processed correctly.

I'd imagine most of these will be calls to GetCertificate / GetCertificates and validation handlers (i.e. there's quite a bit in External API for instance)

src/SFA.DAS.AssessorService.EpaoImporter/PrintProcessCommand.cs
arshedmust: I have updated the code so that we are sanitzing before processing

src/SFA.DAS.AssessorService.EpaoImporter/PrintProcessCommand.cs
arshedmust: This will not be the case anymore since sanitization has now been re introduced. See my previous comment.

src/SFA.DAS.AssessorService.Domain/Consts/CertificateStatus.cs
arshedmust: In most cases we are filtering before we are do anything with certificates with the following flags  Domain.Consts.CertificateStatus.Submitted,
               Domain.Consts.CertificateStatus.Printed,
               Domain.Consts.CertificateStatus.Reprint
So in those cases the Queued will be ignored, which is good since this state a transient state and the certificates should not stay in this state unless something went wrong during printing process.
Once the certificate is in submitted state it does not get modified after that until it gets printed.

arshedmust: Removed the dependency on Queue

arshedmust:
Moved to a new branch ON_1999_V2

PR #820, opened 2019-07-03
On 1999 v1 print function not updating certificates

arshedmust:


src/SFA.DAS.AssessorService.EpaoImporter/PrintProcessCommand.cs
```diff
@@ -47,7 +47,7 @@ public class PrintProcessCommand : ICommand
 
         public async Task Execute()
         {
-            await UploadCertificateDetailsToPinter();
+            await UploadCertificateDetailsToPrinter();
```
davidgouge: Thank you! 😁

PR #821, opened 2019-07-04
ON-1952 Added a column to OrganisationType to record which organisation types are financial exempt.

ChrisJWoodcock:


ChrisJWoodcock:
> Why is src/SFA.DAS.AssessorService.Database/DataSetup/OrganisationType.sql
> attempting to CREATE Table if it already exists? which it should for all environments.
> It should only populate OrganisationType table if it is empty, and otherwise it should simple use a set of update statements to set the Financial Exempt BIT

That is a data setup script for the OrganisationType table - it's creating a temporary table [#OrganisationType] which is populated with the current / to be values which are then merged into the existing table if there are any changes to the current values - it's a common pattern for lookup tables which would be great if the other lookup tables followed as would then be able to create new databases much more easily as well as changing lookup values in a controlled way.

ChrisJWoodcock:
> But the current postdeployment script will fail, because it will try to Create the OrganisationType table and fail, as it is created by the DACPAC. I have no issue with an intelligent script that (re)populates a lookup table on each deploy.

It doesn't fail as the one in the data setup script is a #TemporaryTable; however I think it would dangerous to introduce these types of scripts in isolation so this has been deferred to Tech Debt 2115 and simple updates statements will be used in this ticket.

PR #836, opened 2019-07-16
ON-2012 Record search data for search requests for all search requests; includes IsPrivatelyFunded.

ChrisJWoodcock:


src/SFA.DAS.AssessorService.Domain/Entities/SearchLog.cs
```diff
@@ -8,6 +10,9 @@ public class SearchLog
         public string Surname { get; set; }
         public long Uln { get; set; }
         public DateTime SearchTime { get; set; }
+        public SearchData SearchData { get; set; }
+        [NotMapped]
+        public string CertificateDataJsonString { get => JsonConvert.SerializeObject(SearchData); }
```
ChrisJWoodcock: This is incorrectly named should be SearchDataJsonString

PR #844, opened 2019-07-22
ON-2120 : Add logging for all xxxApiClient errors

martin-lush:


src/SFA.DAS.AssessorService.Application.Api.Client/Clients/ApiClientBase.cs
```diff
@@ -86,41 +86,54 @@ private static HttpRequestException CreateRequestException(HttpRequestMessage fa
...
+                if (result.StatusCode == HttpStatusCode.OK)
                 {
-                    if(!request.RequestUri.IsAbsoluteUri)
-                        message = "Could not find " + request.RequestUri;
-                    else
-                        message = "Could not find " + request.RequestUri.PathAndQuery;
+                    // NOTE: Struct values are valid JSON. For example: 'True'
+                    var json = await result.Content.ReadAsStringAsync();
+                    return await Task.Factory.StartNew<T>(() => JsonConvert.DeserializeObject<T>(json, JsonSettings));
                 }
```
arshedmust: Should update this to use Task.Run instead of Task.Factory.StartNew<T>

martin-hornsby: Keeps it simpler so done it. I'm not sure why GitHub flagged it as new code consider it was there previosuly down below 👇

martin-hornsby: Keeps it simpler so done it. I'm not sure why GitHub flagged it as new code consider it was there previously up above ☝️

src/SFA.DAS.AssessorService.Application.Api.External/Infrastructure/ApiClient.cs
```diff
@@ -60,19 +110,69 @@ protected async Task<T> Get<T>(string uri)
...
+                        catch
+                        {
+                            // safe to ignore any errors
+                        }
+                        _logger.LogError(ex, $"DELETE: HTTP {(int)response.StatusCode} Error getting response from: {uri} - ActualResponse: {actualResponse}");
+                        throw;
+                    }
+                }
+            }
+            catch (HttpRequestException ex)
```
davidgouge: Can this error handling code (that looks the same in each instance) be refactored out?

martin-hornsby: They are HTTP method specific so I think it'll leave a messy refactored method

PR #853, opened 2019-07-31
ON-1909 Report with dynamic pivot table with organisation columns upto the ma…

alandburns:
…ximum number of organistations for any standard in the report.


alandburns:
This change needs 
1) to be checked with Users, that having a dynamic number of columns is allowed
2) if OK, then it needs to be tested

PR #862, opened 2019-08-15
Add PostCodeRegion Table

alandburns:


src/SFA.DAS.AssessorService.Database/Tables/PostCodeRegion.sql
```diff
@@ -0,0 +1,8 @@
+-- PostCode Regions Lookup
+
+CREATE TABLE PostCodeRegion (
+    PostCodePrefix NVARCHAR(2) NOT NULL PRIMARY KEY,
+    Region  NVARCHAR(50)  NOT NULL
+)
```
martin-hornsby: Missing `ON [PRIMARY]` which is consitent with the other tables

alandburns: Only need to use ON [PRIMARY] when you are not using PRIMARY file groups for some of our tables or indexes. It is the Default!
So we should remove it from other tables (which I assume has been added by VS or other "useful" tool) as we are only using PRIMARY.
IF DevOps ever wanted to change this I would leave it to them.

PR #871, opened 2019-08-21
ON-2193 send inviter emails

martin-hornsby:


davidgouge:
Closing as had a re-think about the implementation.

PR #872, opened 2019-08-23
On 2223 data summaries for opportunity finder

alandburns:


alandburns:
> 
> 
> It's a beast! 👍

But it only has to run once a day

PR #876, opened 2019-08-28
ON-2033_anytime_updates_live_contacts_only

ChrisJWoodcock:


src/SFA.DAS.AssessorService.Application.Api.Client/Clients/ContactsApiClient.cs
```diff
@@ -110,6 +110,14 @@ public async Task<List<ContactResponse>> GetAllContactsForOrganisation(string ep
             }
         }
 
+        public async Task<List<ContactResponse>> GetAllContactsWhoCanBePrimaryForOrganisation(string epaoId)
+        {
+            using (var request = new HttpRequestMessage(HttpMethod.Get, $"/api/v1/contacts/get-all-who-can-be-primary/{epaoId}"))
```
arshedmust: Prefer camelcase over name with dash seperators :  'get-all-who-can-be-primary'

PR #878, opened 2019-08-28
ON-2241_Configure_HTTP_Header_from_Security_Audit

alandburns:


src/SFA.DAS.AssessorService.Application.Api.External/web.config
```diff
@@ -3,8 +3,16 @@
...
+        <requestFiltering allowDoubleEscaping="true" removeServerHeader ="true" />
       </security>
+      <httpProtocol>
+        <customHeaders>
+          <remove name="X-Powered-By" />
+          <add name="Cache-Control" value="no-cache, no-store, must-revalidate, pre-check=0, post-check=0, max-age=0, s-maxage=0" />
+          <add name="Pragma" value="no-cache" />
+          <add name="Expires" value="0" />
+        </customHeaders>
+      </httpProtocol>
```
davidgouge: We should be doing these in code rather than web.config really.  
`public static IApplicationBuilder UseSecurityHeaders(this IApplicationBuilder app)
        {
            app.Use(async (context, next) =>
            {
                context.Response.Headers["X-Frame-Options"] = "SAMEORIGIN";
                context.Response.Headers["X-XSS-Protection"] = "1; mode=block";
                context.Response.Headers["X-Content-Type-Options"] = "nosniff";
                context.Response.Headers["Content-Security-Policy"] = "default-src 'self'; img-src 'self' *.google-analytics.com; script-src 'self' 'unsafe-inline' *.googletagmanager.com *.postcodeanywhere.co.uk *.google-analytics.com *.googleapis.com; font-src 'self' data:;";
                context.Response.Headers["Referrer-Policy"] = "strict-origin";
                context.Response.Headers["Cache-Control"] = "no-cache, no-store, must-revalidate";
                context.Response.Headers["Pragma"] = "no-cache";
                await next();
            });

            return app;
        }`

alandburns: Happy for you @davidgouge to change the code. This was Quick Fix but I agree.

PR #891, opened 2019-09-04
ON-2210_dashboard_api_subscriptions_schema

ChrisJWoodcock:


src/SFA.DAS.AssessorService.Database/Tables/Privileges.sql
```diff
@@ -3,6 +3,7 @@
     [UserPrivilege] [nvarchar](120) NOT NULL DEFAULT '',
     [MustBeAtLeastOneUserAssigned] [bit] NOT NULL DEFAULT(0),
     [Description] [nvarchar](MAX) NOT NULL DEFAULT '',
+	[PrivilegeData] [nvarchar](MAX) NOT NULL DEFAULT '',
```
davidgouge: What gets stored in PrivilegeData?

davidgouge: Ahh, just a better way of storing the Description.

PR #903, opened 2019-09-20
On 2095 new organiastion choose standard for apply

arshedmust:


arshedmust:
More work need to be done yet.

PR #904, opened 2019-09-23
ON-2090 : Financial health review dashboard

martin-hornsby:
Added SQL to get Feedback Added FHAs
Added SQL to get Closed FHAs


martin-hornsby:
Alan has confirmed SQL statements are working correctly, HOWEVER we need to start using the new Status values (which should in turn simplify the queries)

PR #921, opened 2019-10-02
On 2310 allow epa os to submit requests for deleted records

arshedmust:


src/SFA.DAS.AssessorService.Application/Handlers/Search/SearchHandler.cs
```diff
@@ -120,7 +121,11 @@ private async Task<List<SearchResult>> Search(SearchQuery request, CancellationT
                         ? new List<SearchResult> {new SearchResult {UlnAlreadyExits = true, Uln = request.Uln, IsPrivatelyFunded = true, IsNoMatchingFamilyName = true } }
                         : new List<SearchResult>();
                 }
-                
+
+                var result = DeletedCertificateResult(certificate, request, likedSurname);
+                if (result.Any())
+                    return result;
+
```
martin-hornsby: Id' probably do it like this to make it more explicit as to what is going on
```
            if (certificate.Status == CertificateStatus.Deleted)
            {
                var result = GetDeletedCertificateResult(certificate, request, likedSurname);
                   if (result.Any())
                       return result;
            }
```

src/SFA.DAS.AssessorService.Application/Handlers/Search/SearchHandler.cs
```diff
@@ -158,6 +172,17 @@ private async Task<List<int>> GetEpaoStandards(Organisation thisEpao)
             return filteredStandardCodes;
         }
 
+        private List<SearchResult> DeletedCertificateResult(Certificate certificate, SearchQuery request, string likedSurname)
```
martin-hornsby: Rename this to GetDeletedCertificateResult

PR #923, opened 2019-10-04
ON-2318 and ON-2319 Apply standard codes not set and set Application Status corectly

arshedmust:
…to 'In Progress' when standard starts as per ON-2319

### DIFF
--- a/src/SFA.DAS.AssessorService.Data/Apply/ApplyRepository.cs
+++ b/src/SFA.DAS.AssessorService.Data/Apply/ApplyRepository.cs
@@ -88,9 +88,11 @@ namespace SFA.DAS.AssessorService.Data.Apply
             using (var connection = new SqlConnection(_configuration.SqlConnectionString))
             {
                 var result = await connection.ExecuteAsync(@"UPDATE Apply
-                                                SET  ApplyData = JSON_MODIFY(JSON_MODIFY(JSON_MODIFY(ApplyData,'$.Apply.StandardReference',@ReferenceNumber),'$.Apply.StandardCode',@StandardCode),'$.Apply.StandardName',@StandardName)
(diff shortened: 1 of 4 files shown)


src/SFA.DAS.AssessorService.Data/Apply/ApplyRepository.cs
```diff
@@ -88,9 +88,11 @@ public async Task UpdateInitialStandardData(UpdateInitialStandardDataRequest sta
             using (var connection = new SqlConnection(_configuration.SqlConnectionString))
             {
                 var result = await connection.ExecuteAsync(@"UPDATE Apply
-                                                SET  ApplyData = JSON_MODIFY(JSON_MODIFY(JSON_MODIFY(ApplyData,'$.Apply.StandardReference',@ReferenceNumber),'$.Apply.StandardCode',@StandardCode),'$.Apply.StandardName',@StandardName)
+                                                SET  ApplyData = JSON_MODIFY(JSON_MODIFY(JSON_MODIFY(ApplyData,'$.Apply.StandardReference',@ReferenceNumber),'$.Apply.StandardCode',@StandardCode),'$.Apply.StandardName',@StandardName),
+                                                    StandardCode = @StandardCode,
+                                                    ApplicationStatus = @ApplicationStatus
```
davidgouge: Not sure why this needs to be an ugly `JSON_MODIFY(JSON_MODIFY(JSON_MODIFY(` sql statement.  Can't we do the ApplyData alterations in code and then just serialize and update the whole json in the table?  So deserialize ApplyData, set the properties, serialize, `UPDATE Apply SET ApplyData = @ApplyData`?
Would be much cleaner / easier to understand / would only need one UpdateApplyData method and the rest would be in the Application / easier to maintain.

arshedmust: Does not need to be, but this is how its done in SQL. Lets create a new story to refactor how we are doing these queries.

arshedmust:
This has been updated as part of another story.

PR #929, opened 2019-10-11
ON-2151 Display the completion status for the learner

ChrisJWoodcock:


src/SFA.DAS.AssessorService.Application/Handlers/Staff/LearnerDetailHandler.cs
```diff
@@ -73,13 +73,36 @@ public async Task<LearnerDetail> Handle(LearnerDetailRequest request, Cancellati
                 OrganisationId = epao.EndPointAssessorOrganisationId,
                 CertificateLogs = logs,
                 FundingModel = learner?.FundingModel,
+                CompletionStatusId = learner?.CompletionStatus,
```
martin-hornsby: Can we keep it as CompletionStatus so it's badly consistent 😁 
And name the other one CompletionStatusDescription or something like that?

PR #931, opened 2019-10-14
On 2326 delete certificate stored proc

alandburns:
This stored proc is ONLY for manual execution (by DevOps), to ensure that a Certificate is correctly 'soft' deleted

### DIFF
--- a/src/SFA.DAS.AssessorService.Database/StoredProcedures/Delete_Certificate.sql
+++ b/src/SFA.DAS.AssessorService.Database/StoredProcedures/Delete_Certificate.sql
@@ -1,0 +1,35 @@
+-- Soft Delete a certificates Record
+CREATE PROCEDURE [Delete_Certificate] 
+@CertificateID Int
+AS
+
+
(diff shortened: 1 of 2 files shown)


src/SFA.DAS.AssessorService.Database/StoredProcedures/Delete_Certificate.sql
```diff
@@ -0,0 +1,37 @@
+-- Soft Delete a certificates Record
+-- this is for manual use only, and not for programatic use
+-- as will need to know the user that is deleting the certificate
+CREATE PROCEDURE [Delete_Certificate] 
+@CertificateID Int
+AS
+
+
+UPDATE certificates 
+SET [status] = 'Deleted', [DeletedBy] = 'manual', [DeletedAt] = GETDATE() , [IsPrivatelyFunded] = 0, [PrivatelyFundedStatus] = NULL
```
davidgouge: `GETUTCDATE()`?  😁

PR #954, opened 2019-11-07
Fixed code to take into account Answer.Value now being an array

davidgouge:


alandburns:
This solution was abandnoned

PR #956, opened 2019-11-11
Fixed issue where CBL was submitting allAreas as an answer and no ans…

davidgouge:
…wers at all if nothing checked


martin-hornsby:
Should be fixing this in ON-2380 via a slightly different approach :)

alandburns:
Has been superceeded by ON-2380

PR #960, opened 2019-12-06
Con 1112 zen desk widget

VasanthaKasirajan3008:


src/SFA.DAS.AssessorService.Web.UnitTests/Helpers/HtmlHelperExtensionsTests.cs
```diff
@@ -0,0 +1,75 @@
...
+    [TestFixture]
+    public class HtmlHelperExtensionsTests
+    {        
+        private string expectedLabelOne;
+        private string expectedLabelTwo;
+        private string expectedOutput;
+        private string actualOutput;      
+
+        [Test]
+        public void WhenICallSetZenDeskLabelsWithMultipleLabelsWithMultipleWordsAndApostrophes_ThenTheOutputIsCorrect()
```
dev4valtech: As the tests are identical, you could have used [TestCases](https://nunit.org/docs/2.5/testCase.html)

src/SFA.DAS.AssessorService.Web/Views/Shared/_ZenDeskWidget.cshtml
```diff
@@ -0,0 +1,22 @@
+﻿@using SFA.DAS.AssessorService.Web.Extensions
+
+@if (ViewBag.HideZenDeskWidget == null || ViewBag.HideZenDeskWidget == false)
```
dev4valtech: Where is `HideZenDeskWidget` being set? Couldn't the existence of ZenDeskLabel be used as the indicator instead?

src/SFA.DAS.AssessorService.Web/Extensions/HtmlHelperExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+        {            
+            return new HtmlString($"<script type=\"text/javascript\">zE('webWidget', 'helpCenter:setSuggestions', {{  search: '{suggestion}' }});</script>");
+        }
+
+        public static IHtmlContent SetZenDeskLabels(this IHtmlHelper html, params string[] labels)
+        {
+            var apiCallString = "<script type=\"text/javascript\">zE('webWidget', 'helpCenter:setSuggestions', { labels: [";
+
+            var first = true;
+            foreach (var label in labels)
```
dev4valtech: You could use String.Join instead of the foreach construct

src/SFA.DAS.AssessorService.Web.UnitTests/Helpers/HtmlHelperExtensionsTests.cs
VasanthaKasirajan3008: Added TestCaseSource

src/SFA.DAS.AssessorService.Web/Views/Shared/_ZenDeskWidget.cshtml
VasanthaKasirajan3008: Removed HideZenDeskWidget , Its not required ... we need to display Help Widget all the pages

PR #966, opened 2020-01-06
CON-1217 Added a settings table to assessor database and endpoints to assessor API

ChrisJWoodcock:


src/SFA.DAS.AssessorService.Data/SettingRepository.cs
```diff
@@ -0,0 +1,58 @@
...
+                    Name = name
+                },
+                transaction: _unitOfWork.Transaction);
+
+            return value;
+        }
+
+        public async Task SetSetting(string name, string value)
+        {
+            const string sql = "SELECT [Name] FROM [Settings] WHERE [Name] = @Name";
```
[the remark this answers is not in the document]
ChrisJWoodcock: It would be easier but about 95% of all the code looks similar to this currently - so I think it's best left looking as consistent as possible - I am adding stored procedure where they are overwhelming necessary though

src/SFA.DAS.AssessorService.Application/Handlers/Settings/SetSettingHandler.cs
```diff
@@ -15,10 +15,30 @@ public SetSettingHandler(ISettingRepository settingRepository)
             _settingRepository = settingRepository;
         }
 
-        public async Task<Unit> Handle(SetSettingRequest request, CancellationToken cancellationToken)
+        public async Task<SetSettingResult> Handle(SetSettingRequest request, CancellationToken cancellationToken)
         {
-            await _settingRepository.SetSetting(request.Name, request.Value);
-            return Unit.Value;
+            var result = new SetSettingResult();
+            if(request.Name.Length == 0 || request.Name.Length > 50)
```
cofaulco: The usual pattern is to have a separate validator to validate the request.

src/SFA.DAS.AssessorService.Data/SettingRepository.cs
```diff
@@ -44,15 +44,17 @@ public async Task SetSetting(string name, string value)
...
-                await _unitOfWork.Connection.ExecuteAsync(
+         
+            await _unitOfWork.Connection.ExecuteAsync(
                 "UPDATE [Settings] SET [Value] = @value " +
                 "WHERE [Name] = @name",
                 param: new { value, name },
                 transaction: _unitOfWork.Transaction);
-            }
+            
+            return false;
```
cofaulco: the caller of this class would have no idea what a true result means without drilling down and inspecting the code.

src/SFA.DAS.AssessorService.Application/Handlers/Settings/SetSettingHandler.cs
ChrisJWoodcock: Ok - I see that now

src/SFA.DAS.AssessorService.Data/SettingRepository.cs
ChrisJWoodcock: Refactored to externalise the result

ChrisJWoodcock:
The latest PEN test has been completed it has shown the same LOW risk elements raised on the previous test Jan 2019 - these low risk elements will not be fixed.
[https://skillsfundingagency.atlassian.net/wiki/download/attachments/125142558/Internal%20API%20Assessor%20Pen%20Test%202.pdf?version=1&modificationDate=1580741273082&cacheVersion=1&api=v2]

PR #970, opened 2020-01-10
CON-1094 Remove the redundant provider events version of epao data sync

ChrisJWoodcock:
Replaced by the assessor functions version


ChrisJWoodcock:
This is the removal of a redundant function app - it will need some input from dev ops to remove it from the build and pipelines.

PR #973, opened 2020-01-16
ON-2386 charitable incorporated organisations

martin-hornsby:


src/SFA.DAS.AssessorService.Application.Api/Infrastructure/CompaniesHouseApiClient.cs
```diff
@@ -50,8 +50,14 @@ public async Task<bool> IsCompanyActivelyTrading(string companyNumber)
 
             if (company != null)
             {
-                isTrading = "active".Equals(company.Status, StringComparison.InvariantCultureIgnoreCase) &&
-                            company.DissolvedOn == null && company.IsLiquidated != true;
+                if(company.Status is null && "charitable-incorporated-organisation".Equals(company.Type, StringComparison.InvariantCultureIgnoreCase))
```
David-B-Read: Should this string be a constant?

martin-hornsby: yes but the code below is done in the same way so keeping it consistent ;)

PR #979, opened 2020-01-22
CON-1134 Suppressing only those standards which are explicitly 'integrated degree'

ChrisJWoodcock:


src/SFA.DAS.AssessorService.Database/Functions/Scalar-valued Functions/OppFinder_Is_Approved_StandardStatus.sql
```diff
@@ -6,8 +6,8 @@ RETURNS BIT
 AS
 BEGIN
 	IF (ISNULL(JSON_VALUE(@StandardData, '$.IfaStatus'), '') = 'Approved for delivery')
-	   AND -- if there is any value in IntegratedDegree it is not a valid Approved standard
-	   (ISNULL(JSON_VALUE(@StandardData, '$.IntegratedDegree'), '') = '')
+	   AND -- standards which are integrated degrees are not valid approved standards
```
cofaulco: might be good to be a little more specific here just because looking back we may forget the context.

> Standards which have the values of "**integrated degree**" in the IntegratedDegree field are not valid approved standards

PR #981, opened 2020-01-22
On 2375 newtonsoft error

cofaulco:


cofaulco:
> we don't want to comment out StandardCollationFlow

Fixed

PR #984, opened 2020-01-23
(re)Fix gitignore

alandburns:


ChrisJWoodcock:
The purpose of this change is not clear; it is correct to ignore .min.js and .min.css files in .gitignore; however there are lots of checked in minified files (which may be necessary if the un-minified files are not available); which may force a more selective ignore policy; this is a revert of code which was committed 6 months ago and hasn't been reported as an issue since; a tech debt will be required to investigate and correct the bundling in the project.

.gitignore
```diff
@@ -276,8 +276,12 @@ src/Tools/PhantomJS/tools/phantomjs/ChangeLog
 src/SFA.DAS.AssessorService.Web.Specflow.Tests/Project/**/*.*
 
 # Minified css and js
```
ChrisJWoodcock: Since this is being done for an unusual reason - because there are included minified files in the repository and a blanket ignore is implicated in a build time bundling issue - a better comment would appropriate here

PR #998, opened 2020-03-25
CON-1612 bug fix

shomavg:


src/SFA.DAS.AssessorService.Web/Controllers/Apply/ApplicationController.cs
```diff
@@ -897,7 +897,7 @@ private List<Answer> GetAnswersFromForm(Page page)
             List<Answer> answers = new List<Answer>();
 
             // These are special in that they drive other things and thus should not be deemed as an answer
-            var exludedInputs = new List<string> { "postcodeSearch", "checkAll" };
+            var exludedInputs = new List<string> { "postcodeSearch", "checkAll", "formAction", "redirectAction", "summaryLink" };
```
ChrisJWoodcock: You could take the opportunity to spell this correctly

ChrisJWoodcock:
This pull request is being abandoned as I have merged it into a mess

PR #1003, opened 2020-04-17
Con 1689 co browsing

VasanthaKasirajan3008:


VasanthaKasirajan3008:
> Has the config been done?

yes, https://github.com/SkillsFundingAgency/das-employer-config/pull/549

PR #1005, opened 2020-04-21
CON-1666 Loading options from IFATE

ChrisJWoodcock:


src/SFA.DAS.AssessorService.Application.Api/Services/StandardService.cs
```diff
@@ -76,6 +76,22 @@ public async Task<StandardCollation> GetStandard(string referenceNumber)
...
+        {
+            List<Option> options = null;
+
+            try
+            {
+                options = await _standardRepository.GetOptions(stdCode);
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, $"OPTION: Failed to get for options for stdcode: {stdCode}");
```
Stephek, line 89: Should the error be propagated rather than swallowed here?

src/SFA.DAS.AssessorService.Application.Api/Services/StandardService.cs
```diff
@@ -203,6 +222,22 @@ private StandardCollation MapDataToStandardCollation(int standardId, IfaStandard
             };
         }
 
+        private List<string> MapIfaStandardOptions(IfaStandard ifaStandard)
```
Stephek, line 225: Could this method be put on the IfaStandard object?  

e.g.  ifaSrtandard.GetOptionsList()

src/SFA.DAS.AssessorService.Application/Interfaces/IStandardService.cs
```diff
@@ -12,6 +12,7 @@ public interface IStandardService
 
         Task<StandardCollation> GetStandard(int standardId);
         Task<StandardCollation> GetStandard(string referenceNumber);
+        Task<List<Option>> GetOptions(int stdCode);
```
Stephek, line 15: Shouldn't this method be on a IOptionsService/repository interface since it is returning Options rather than Standards.

Or it should return a list of Standards for the stdCode.  Then the caller can extract the Options as required using a linq query.

src/SFA.DAS.AssessorService.Application/Interfaces/IStandardRepository.cs
```diff
@@ -18,6 +18,8 @@ public interface IStandardRepository
         Task<List<StandardNonApprovedCollation>> GetStandardNonApprovedCollations();
         Task<StandardNonApprovedCollation> GetStandardNonApprovedCollationByReferenceNumber(string referenceNumber);
 
+        Task<List<Option>> GetOptions(int stdCode);
```
Stephek, line 21: See comment below.

src/SFA.DAS.AssessorService.Data.IntegrationTests/Handlers/StandardCollationHandler.cs
```diff
@@ -10,17 +10,36 @@ public class StandardCollationHandler
...
+
+            var sqlToInsertOption =
+                "INSERT INTO [dbo].[Options]" +
+                    "([StdCode]" +
+                    ",[OptionName]" +
+                    ",[IsLive])" +
+                "VALUES" +
+                    "(@stdCode" +
+                    ",@optionName" +
+                    ",@isLive);";
```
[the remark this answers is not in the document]
ChrisJWoodcock: This is only test data to back some integration tests

src/SFA.DAS.AssessorService.Application/Interfaces/IStandardService.cs
ChrisJWoodcock: I agree - I attempted to do, but I moved it from the Certificate repository to here so that I could re-used the SQL as that was a definite improvement - attempting to move it to a new separate repository set of a cascade of changes which quickly exceeded the scope of the ticket - there is a separate Tech debt for untangling the mishmash of references coming soon....

src/SFA.DAS.AssessorService.Application/Interfaces/IStandardRepository.cs
ChrisJWoodcock: As above explanation

src/SFA.DAS.AssessorService.Application.Api/Services/StandardService.cs
ChrisJWoodcock: Agree - done

src/SFA.DAS.AssessorService.Application.Api/Services/StandardService.cs
ChrisJWoodcock: Remove the whole method as it could be refactored away

src/SFA.DAS.AssessorService.Application/Interfaces/IStandardService.cs
ChrisJWoodcock: Refactored the single caller to avoid using this method and removed the method.

src/SFA.DAS.AssessorService.Application/Interfaces/IStandardRepository.cs
ChrisJWoodcock: This has been moved from the certificate repository as options have more to do standards that certificates; it is possible for the caller of this method (certificate creation) to get its options from its standard (making an Options repository unnecessary) but currently this causes a large amount of cascaded changes which is outside of the scope of this ticket - there is a Tech Debt ticket to start to unravel this stuff after which could make sensible changes like that in the future.

src/SFA.DAS.AssessorService.Data.IntegrationTests/Handlers/StandardCollationHandler.cs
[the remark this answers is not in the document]
cofaulco: my bad. Ignore the point

PR #1014, opened 2020-06-03
CON-1834 track print status

ChrisJWoodcock:
Added endpoints for batch log sent to printer and certificate print status, requires unit tests adding


src/SFA.DAS.AssessorService.Application/Handlers/Certificates/UpdateCertificatesPrintStatusHandler.cs
```diff
@@ -0,0 +1,79 @@
...
+                if (!notFoundBatches.Contains(certificatePrintStatus.BatchNumber))
+                {
+                    if(await _mediator.Send(new GetForBatchNumberBatchLogRequest { BatchNumber = certificatePrintStatus.BatchNumber }) == null)
+                    {
+                        validationResult.Errors.Add(new ValidationErrorDetail(nameof(request.CertificatePrintStatuses), $"The batch number {certificatePrintStatus.BatchNumber} was not found.", ValidationStatusCode.NotFound));
+                        notFoundBatches.Add(certificatePrintStatus.BatchNumber);
+                    }
+                }
+
+                if (!notFoundBatches.Contains(certificatePrintStatus.BatchNumber))
```
cofaulco, line 46: same if statement as above, this code could be a continuation of the above block with a `continue;` after the `notFoundBatches.Add()`

src/SFA.DAS.AssessorService.Application/Handlers/Certificates/UpdateCertificatesPrintStatusHandler.cs
```diff
@@ -0,0 +1,79 @@
...
+                    }
+                    else
+                    {
+                        // when the certificate batch number is not set then a reprint has been requested but not sent to printer
+                        // any print status update would be for a prior batch number and would not update the certificate status
+                        var changesCertificateStatus = certificatePrintStatus.BatchNumber >= (certificate.BatchNumber ?? int.MaxValue) &&
+                            certificatePrintStatus.StatusChangedAt > certificate.LatestChange().Value &&
+                            certificate.Status != CertificateStatus.Deleted;
+
+                        if (!validPrintStatus.Contains(certificatePrintStatus.Status))
```
cofaulco, line 61: Could this check be done first thing within this loop to shortcut processing for an invalid status. Would save multiple roundtrips to the DB

src/SFA.DAS.AssessorService.Data/CertificateRepository.cs
```diff
@@ -184,6 +187,14 @@ private bool CheckCertificateData(Certificate certificate, string lastName)
             return certificate;
         }
 
+        public async Task<Certificate> GetCertificate(string certificateReference)
+        {
+            var certificate = await
+               _context.Certificates
```
cofaulco, line 193: Is there a reason why in the BatchLogRepository we got rid of the EF context and replaced with Dapper SQL extensions but here we are keeping the EF Context?

ChrisJWoodcock: The whole application was written as a mixture of Dapper and EF; that's bad enough but having repositories which are themselves a mixture of two data access techniques is awful - If I'm changing a repository and the majority is already Dapper I'm converting to Dapper if its simple enough to do - ones such as certificates and contact which are large and almost entirely EF I'm leaving as EF - it would be a larger Tech Debt to convert them.

PR #1015, opened 2020-06-05
CON-1834 Added schema to support tracking the print status of certificates.

ChrisJWoodcock:


src/SFA.DAS.AssessorService.Database/PreDeploymentScripts/remove_certificatelogs_wasrejected.sql
```diff
@@ -0,0 +1,14 @@
+﻿-- This script was extracted from the PreDeploymentScript it has currently worked its way 
+-- into PROD and so has been removed from the PreDeploymentScript
```
cofaulco, line 2: Comments seem to suggest that this script has been removed from PreDeployment however it seems it is still there?

PR #1019, opened 2020-06-15
Correct SQL Casing

CodeWolf86:


CodeWolf86:
SonarCloud failed due to updating the SQL statements to fix the warnings in visual studio. It's not duplicated code and therefore merging on the basis that the Sonar Block is a redherring.

PR #1021, opened 2020-06-24
CON-2002 : EPAO - Certificate address

VasanthaKasirajan3008:


src/SFA.DAS.AssessorService.Web/Views/Certificate/Check.cshtml
```diff
@@ -272,7 +272,14 @@ else
                             Department
                         </dt>
                         <dd class="govuk-summary-list__value">
-                            @Model.Dept
+                            @if (string.IsNullOrEmpty(Model.Dept))
```
shomavg, line 275: @Model?

VasanthaKasirajan3008: done

PR #1032, opened 2020-07-21
Con 2061 transfer recipient from email templates

VasanthaKasirajan3008:


src/SFA.DAS.AssessorService.Data/EMailTemplateQueryRepository.cs
```diff
@@ -1,24 +1,30 @@
...
         {
-            _assessorDbContext = assessorDbContext;
         }
 
-        public async Task<EMailTemplate> GetEmailTemplate(string templateName)
+        public async Task<EmailTemplateSummary> GetEmailTemplate(string templateName)
         {
-            return await _assessorDbContext.EMailTemplates.FirstOrDefaultAsync(q => q.TemplateName == templateName);
+            var sql = @"SELECT ET.Id, ET.TemplateName, ET.TemplateId, ETR.Recipients, ET.RecipientTemplate
+                    FROM [dbo].[EMailTemplates] ET join [dbo].[EmailTemplatesRecipients] ETR
```
ChrisJWoodcock, line 18: should use inner join here explicitly (join is the same thing but inner is clearer)

src/SFA.DAS.AssessorService.Data/EMailTemplateQueryRepository.cs
```diff
@@ -1,24 +1,30 @@
...
+                    ON ET.Id = ETR.EmailTemplateId
+                    WHERE ET.TemplateName = @templateName";
+
+            var results = await _unitOfWork.Connection.QueryAsync<EmailTemplateSummary>(
+                sql,
+                param: new { templateName },
+                transaction: _unitOfWork.Transaction);
+
+            return results.FirstOrDefault();
```
ChrisJWoodcock, line 27: The underlying table structure may include more than one recipient row for an email template - there is nothing to stop that, so taking the first is a potential bug waiting to happen - suggest you add a unique key to the table to prevent that ever happening

src/SFA.DAS.AssessorService.Application/Handlers/EmailHandlers/SendEmailHandler.cs
```diff
@@ -18,60 +19,39 @@ public class SendEmailHandler : IRequestHandler<SendEmailRequest>
...
+            }
+            else if (emailTemplateSummary != null && emailTemplateSummary.Recipients != string.Empty)
             {
-                var personalisationTokens = GetPersonalisationTokens(message.Tokens);
-                await SendEmailViaNotificationsApi(message.Email, emailTemplate.TemplateId, emailTemplate.TemplateName, personalisationTokens);
-
-                if(emailTemplate.RecipientTemplate != null)
-                    await SendEmailToRecipients(emailTemplate.RecipientTemplate, personalisationTokens);
-
+                await SendEmail(emailTemplateSummary, personalisationTokens);
```
ChrisJWoodcock, line 43: SendEmailToTemplateRecipients may make this clearer as it has some pluralisation in the name

src/SFA.DAS.AssessorService.Database/Tables/EmailTemplatesRecipients.sql
```diff
@@ -0,0 +1,20 @@
...
+	[CreatedBy] [nvarchar](256) NOT NULL,
+	[UpdatedAt] [datetime2](7) NULL,
+	[UpdatedBy] [nvarchar](256) NULL,
+	[DeletedAt] [datetime2](7) NULL,
+	[DeletedBy] [nvarchar](256) NULL,
+	CONSTRAINT [PK_EmailTemplatesRecipients] PRIMARY KEY ([Id]),
+) ON [PRIMARY] 
+GO
+
+ALTER TABLE [dbo].[EmailTemplatesRecipients]
```
ChrisJWoodcock, line 16: The unique key needs to here as well or it will be removed....

src/SFA.DAS.AssessorService.Database/PostDeploymentScripts/con-2061-clear-entries-recipienttemplate.sql
```diff
@@ -0,0 +1,18 @@
+BEGIN TRANSACTION
```
ChrisJWoodcock, line 1: This file should be named something like 'con-2061-transfer-recipient-template-to-recipients'

src/SFA.DAS.AssessorService.Database/PostDeploymentScripts/con-2061-email-templates-recipients.sql
```diff
@@ -0,0 +1,10 @@
+BEGIN TRANSACTION
```
ChrisJWoodcock, line 1: This file should be named something like 'con-2061-create-email-template-recipients' and then it should clear the values in the [EmailTemplates].Recipients column when it has finished

src/SFA.DAS.AssessorService.Database/PostDeploymentScripts/con-2061-clear-entries-recipienttemplate.sql
VasanthaKasirajan3008: done

src/SFA.DAS.AssessorService.Database/PostDeploymentScripts/con-2061-email-templates-recipients.sql
VasanthaKasirajan3008: done

src/SFA.DAS.AssessorService.Data/EMailTemplateQueryRepository.cs
VasanthaKasirajan3008: done

src/SFA.DAS.AssessorService.Database/SFA.DAS.AssessorService.Database.sqlproj
```diff
@@ -142,6 +142,9 @@
     <None Include="PostDeploymentScripts\con-1834-add_certificatebatchlogs_data.sql" />
     <None Include="PostDeploymentScripts\qual-39-remove-contact-roles.sql" />
     <Build Include="Tables\EmailTemplatesRecipients.sql" />
+    <None Include="PostDeploymentScripts\con-2061-clear-recipients-from-email-templates.sql" />
```
ChrisJWoodcock, line 145: Will this work in this order - if you clear them what would you transfer?

ChrisJWoodcock (approved):

PR #1034, opened 2020-07-22
Con 2074 display print process batch status

hkf-tech:


src/SFA.DAS.AssessorService.Application/Interfaces/IStaffCertificateRepository.cs
```diff
@@ -19,14 +21,17 @@ public interface IStaffCertificateRepository
 
     public class StaffReposBatchSearchResult
```
cofaulco, line 22: why do these classes live in this Repository interface file?

src/SFA.DAS.AssessorService.Data/CertificateRepository.cs
```diff
@@ -193,7 +204,8 @@ private bool CheckCertificateData(Certificate certificate, string lastName, Date
 
         public async Task<List<Certificate>> GetCompletedCertificatesFor(long uln)
         {
-            return await _context.Certificates.Where(c => c.Uln == uln && (c.Status == CertificateStatus.Reprint || c.Status == CertificateStatus.Printed || c.Status == CertificateStatus.Submitted || c.Status == CertificateStatus.ToBeApproved))
+            var statuses = new[] { CertificateStatus.Submitted, CertificateStatus.ToBeApproved }.Concat(CertificateStatus.PrintProcessStatus).ToList();
```
cofaulco, line 207: are these collectively known as "Completed statuses"?

src/SFA.DAS.AssessorService.Data/CertificateRepository.cs
```diff
@@ -399,45 +415,93 @@ public async Task<Certificate> UpdateProviderName(Guid id, string providerName)
...
+            var certificateBatchLog =
+                await _context.CertificateBatchLogs.FirstOrDefaultAsync(
+                    q => q.CertificateReference == certificate.CertificateReference && q.BatchNumber == batchNumber);
 
-            foreach (var certificateStatus in updateCertificatesBatchToIndicatePrintedRequest.CertificateStatuses)
+            if (certificateBatchLog != null)
             {
-                var certificate =
-                    await _context.Certificates.FirstAsync(q => q.CertificateReference == certificateStatus.CertificateReference);
+                if (changesCertificateStatus)
```
cofaulco, line 447: should this check be done outside of the repository? I wouldn't expect to check whether to change the status inside a method specifically for updating the status

ChrisJWoodcock: I don't understand the question here - are you saying that the check should be done inside the repository method or outside? currently the check is done outside which to me seems correct.

src/SFA.DAS.AssessorService.Data/CertificateRepository.cs
ChrisJWoodcock: Recently I have been trying to collect together status sets in the certificate status class - it is possible that this set of status represents a completed status, it is not used else where in this combination - I'm not sure if that does make it a collective status or not? I'm thinking that where it is used at least twice it could be promoted

src/SFA.DAS.AssessorService.Application/Interfaces/IStaffCertificateRepository.cs
ChrisJWoodcock: There were always there - they should be moved to the DTO area in the domain library

src/SFA.DAS.AssessorService.Application.UnitTests/Handlers/BatchLogs/PrintedBatchLogHandlerTests/When_called_and_batch_does_exist.cs
```diff
@@ -0,0 +1,108 @@
...
+        public void Then_validation_response_is_valid_true()
+        {
+            _response.IsValid.Should().Be(true);
+            _response.Errors.Count.Should().Be(0);
+        }
+
+        [Test]
+        public void Then_certificate_update_print_status_message_is_sent()
+        {
+            _mediator.Verify(r => r.Send(It.Is<UpdateCertificatesPrintStatusRequest>(t => t.CertificatePrintStatuses.Count == 2), It.IsAny<CancellationToken>()));
```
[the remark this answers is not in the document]
ChrisJWoodcock: Ok

src/SFA.DAS.AssessorService.Data/CertificateRepository.cs
cofaulco: I was just thinking that the variable could be named as such for now to make it apparent. I made the connection based on the method name but would maybe be better to be explicit

src/SFA.DAS.AssessorService.Data/CertificateRepository.cs
cofaulco: what I mean is, we shouldn't need to pass in the changesCertificateStatus flag and check it.
This method should only be called if ChangesCertificateStatus = true

ChrisJWoodcock: The method changes the print status; which is a combination of certificate status, batch log status and certificate logs; there is business logic for whether the certificate status is changed; simple logic for batch log status and the certificate logs are always updated, so I think it's correct that the flag is passed from the handler to the repository for updating the certificate status as the handler has the unit tests.

src/SFA.DAS.AssessorService.Data/CertificateRepository.cs
ChrisJWoodcock: Ok - will rename the variable for clarity

src/SFA.DAS.AssessorService.Data/CertificateRepository.cs
cofaulco: ok, I think I see where you are coming from. Resolving this

PR #1039, opened 2020-08-05
Con 1568 config changes

hkf-tech:


src/SFA.DAS.AssessorService.Settings/WebConfiguration.cs
```diff
@@ -14,26 +12,20 @@ public class WebConfiguration : IWebConfiguration
...
         [JsonRequired] public string SqlConnectionString { get; set; }
 
-        public string SpecflowDBTestConnectionString { get; set; }
-
         [JsonRequired] public string SessionRedisConnectionString { get; set; }
-        [JsonRequired] public AuthSettings StaffAuthentication { get; set; }
+        
         [JsonRequired] public ClientApiAuthentication QnaApiAuthentication { get; set; }
         [JsonRequired] public string ServiceLink { get; set; }
         [JsonRequired] public DfeSignInConfig DfeSignIn { get; set; }
```
ChrisJWoodcock, line 27: When EPAO originally written the DfeSignIn was used for authentication; then a new service the [das-login-service] was built instead - there was talk about this being temporary - going back to the DfeSignIn or using an entirely different government sign-in; I think the use of the LoginService will be long term and this part of the configuration should be renamed to appropriately.

src/SFA.DAS.AssessorService.Settings/IWebConfiguration.cs
```diff
@@ -2,24 +2,19 @@
 {
     public interface IWebConfiguration
     {
-        AuthSettings Authentication { get; set; }
         ApiAuthentication ApiAuthentication { get; set; }
         AzureApiAuthentication AzureApiAuthentication { get; set; }
         ClientApiAuthentication ClientApiAuthentication { get; set; }
```
ChrisJWoodcock, line 7: There are several uses of the ClientApiAuthentication - the use of this one is for the Assessor Web to connect to the Assessor Api; so this one should be called AssessorApiAuthentication

PR #1043, opened 2020-08-21
CON-2206- Update certificate confirmation page

ben1stone:


src/SFA.DAS.AssessorService.Web/Views/Certificate/Confirmation.cshtml
```diff
@@ -21,7 +21,7 @@
                         <br /><strong>@Model.GivenNames @Model.FamilyName</strong>
                     </div>
                     <div>
-                         @if (Model.SelectedGrade == CertificateGrade.Fail)
+                        @if (Model.SelectedGrade == CertificateGrade.Fail)
```
ChrisJWoodcock, line 24: The AC's state that the screen will change when it is NOT a fail - I'm not sure that is correct - it should probably change for both a pass and a fail - but the fail case has slightly different text - the implementation seems more correct than the AC's but they should match for testing - I think the AC's should state what happens for a pass and for a fail

src/SFA.DAS.AssessorService.Web/Views/Certificate/Confirmation.cshtml
```diff
@@ -35,10 +35,16 @@
                 </div>
             </div>
 
+            <h2 class="govuk-heading-m">What happens next</h2>
```
ChrisJWoodcock, line 38: The AC's state the colour is changing - I don't see a change for this unless it's CDN supplied? and I didn't think this was using a CDN?

src/SFA.DAS.AssessorService.Web/Views/Certificate/Confirmation.cshtml
```diff
@@ -35,10 +35,16 @@
...
+
+            <p class="govuk-body"> We'll send the apprentice's certificate to the address you entered. You can <a class="govuk-link" asp-controller="CertificateHistory" asp-action="Index">check the status of the certificate</a> in the 'Completed assessments' section of your account.</p>
```
ChrisJWoodcock, line 40: Mostly this page uses a localizer to supply the text - I think it should either continue to mostly do so - for the new text or not do that at all and convert back to inline text completely.

My preference would be to convert it to inline text as mostly the rest of the application does not use localizers - as EFSA is only for England...

ChrisJWoodcock (changes requested):

src/SFA.DAS.AssessorService.Web/Views/Certificate/Confirmation.cshtml
cofaulco: AC's updated, should not show new content for Fail grade however should display the new link

src/SFA.DAS.AssessorService.Web/Views/Certificate/Confirmation.cshtml
cofaulco: Upon investigation, the new GDS colour has already been implemented.

src/SFA.DAS.AssessorService.Web/Views/Certificate/Confirmation.cshtml
cofaulco: As per @ChrisJWoodcock  recommendation, remove Localizer for this View.

src/SFA.DAS.AssessorService.Web/Views/Certificate/Confirmation.cshtml
```diff
@@ -35,16 +35,22 @@
...
+                    We'll send the apprentice's certificate to the address you entered. You can <a class="govuk-link" asp-controller="CertificateHistory" asp-action="Index">check the status of the certificate</a> in the 'Completed assessments' section of your account.
+                </p>
+            }
             <h2 class="govuk-heading-m">@Localizer["DoNextHeader"]</h2>
             <ul class="govuk-list">
                 <li><a class="govuk-link" asp-controller="Search" asp-action="Index">@Localizer["DoNextSearch"]</a></li>
                 <li><a class="govuk-link" href="https://www.smartsurvey.co.uk/s/certification-transaction-complete/" target="_blank">@Localizer["DoNextFeedback"]</a></li>
-                <li>or</li>
-                <li><a class="govuk-link" asp-controller="Dashboard" asp-action="Index">Return to account home</a></li>
+                @if (Model.SelectedGrade != CertificateGrade.Fail)
```
ChrisJWoodcock, line 49: I don't think AC4 required any changes - it was a test scenario to show that the link at the end applied to both the Fail and the Pass - so this is now broken

ChrisJWoodcock (changes requested):
The removal of the Localizer on this page has not been done

src/SFA.DAS.AssessorService.Web/Views/Certificate/Confirmation.cshtml
ChrisJWoodcock: @shomavg - you haven't done this yet

ChrisJWoodcock (approved):

PR #1054, opened 2020-09-14
CON-1888 Removed the commented out duplicated certificate removal code and tables were these were being stored.

ChrisJWoodcock:


cofaulco (dismissed):
Could do with another task to clear our the post deployment script we have added to clear out the tables after this has been released

cofaulco (approved):

PR #1055, opened 2020-09-17
CON-2142 raise error when exception

hkf-tech:


src/SFA.DAS.AssessorService.Application.Api/Controllers/OppFinder/OppFinderController.cs
```diff
@@ -66,6 +66,7 @@ public async Task UpdateStandardSummary([FromBody] UpdateStandardSummaryRequest
             catch(Exception ex)
             {
                 _logger.LogError(ex, "Request to update standard summary failed");
+                throw new Exception("Something went wrong, Please contact support");
```
ChrisJWoodcock, line 69: I think the message thrown should be the same as the log message - as it is the callers responsibility to log a 'friendly' message if it wants too.

ChrisJWoodcock (changes requested):

src/SFA.DAS.AssessorService.Application.Api/Controllers/OppFinder/OppFinderController.cs
hkf-tech: Changed now it will return the exception thrown.

ChrisJWoodcock: I meant something like this - throw an exception with what happened just don't recommend what the caller should do about it

try	
{	
    _logger.LogInformation($"Received request to update standard summary");	      
     await _mediator.Send(request);
}	
catch(Exception ex)	
{	
    _logger.LogError(ex, "Request to update standard summary failed");	
    throw new Exception("Request to update standard summary failed");	
}

ChrisJWoodcock (approved):

PR #1057, opened 2020-09-28
Dasd 8153 use managed identity

smneal:
- Updated code to support MI connectivity.
- Does the API user also need adding to the ass-sbox-as databases as well?
- ExternalApiDataSync is probably being moved to the das-assessor-functions repo so changes would now be redundant here.

**To Do**
- [ ] Add das-ENV-assapi-as user to PROD and MO. Already added to AT, TEST, TEST2, PREPROD and DEMO.
- [ ] SqlConnection needs updating to use the MI connection string in the das-assessor-service release pipeline.


src/SFA.DAS.AssessorService.Application.Api/StartupConfiguration/Startup.cs
```diff
@@ -57,6 +60,11 @@ public Startup(IHostingEnvironment env, IConfiguration config, ILogger<Startup>
                 UseSandbox = "yes".Equals(config["UseSandboxServices"], StringComparison.InvariantCultureIgnoreCase);
             }
 
+            if (!_env.IsDevelopment())
+            {
+                _azureServiceTokenProvider = new AzureServiceTokenProvider();
```
adam230594, line 65: I wonder if theres any implications of generating the bearer token in startup. Can you confirm there are no issues with the app an hour after deployment?

src/SFA.DAS.AssessorService.Database/Tables/EmailTemplatesRecipients.sql
```diff
@@ -16,7 +16,7 @@ GO
 ALTER TABLE [dbo].[EmailTemplatesRecipients] 
 ADD CONSTRAINT [FK_EmailTemplates_EmailTemplatesRecipients] 
 FOREIGN KEY([EmailTemplateId])
-REFERENCES [dbo].[EmailTemplates] ([Id]);
+REFERENCES [dbo].[EMailTemplates] ([Id]);
```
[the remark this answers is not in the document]
smneal: Yes. The reference is wrong believe it or not.

src/SFA.DAS.AssessorService.Application.Api/StartupConfiguration/Startup.cs
```diff
@@ -220,7 +228,7 @@ private IServiceProvider ConfigureIOC(IServiceCollection services)
                 option.UseSqlServer(sqlConnectionString, options => options.EnableRetryOnFailure(3));
 
                 config.For<AssessorDbContext>().Use(c => new AssessorDbContext(option.Options));
-                config.For<IDbConnection>().Use(c => new SqlConnection(sqlConnectionString));
+                config.For<IDbConnection>().Use(c => ManagedIdentitySqlConnection.GetSqlConnection(sqlConnectionString, _azureServiceTokenProvider));
```
NickGraham101, line 231: Following on from Adam's [comment](https://github.com/SkillsFundingAgency/das-assessor-service/pull/1057/files#r496557271), we need to understand and test what uses this db connection service.  If it's something like NServiceBus issues with token expiry might not be surfaced just by waiting 1 hour

src/SFA.DAS.AssessorService.ExternalApiDataSync/Helpers/ManagedIdentitySqlConnection.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Microsoft.Azure.Services.AppAuthentication;
```
NickGraham101, line 1: This duplicates src/SFA.DAS.AssessorService.Data/Helpers/ManagedIdentitySqlConnection.cs.  Can the duplication be removed by aligning the target frameworks so you can reference the helper class in SFA.DAS.AssessorService.Data rather than duplicating code?

src/SFA.DAS.AssessorService.ExternalApiDataSync/Startup/Bootstrapper.cs
```diff
@@ -1,8 +1,12 @@
-﻿using Microsoft.Azure.WebJobs;
+﻿using Microsoft.Azure.Services.AppAuthentication;
+using Microsoft.Azure.WebJobs;
 using Microsoft.Extensions.Logging;
+using SFA.DAS.AssessorService.ExternalApiDataSync.Helpers;
```
NickGraham101, line 4: See [this comment](https://github.com/SkillsFundingAgency/das-assessor-service/pull/1057/files#r517265542) re duplication

src/SFA.DAS.AssessorService.ExternalApiDataSync/Startup/Bootstrapper.cs
```diff
@@ -31,7 +43,7 @@ public static void StartUp(ILogger functionLogger, ExecutionContext context)
 
                 configure.For<IAggregateLogger>().Use(_logger).Singleton();
                 configure.For<IWebConfiguration>().Use(configuration).Singleton();
-                configure.For<IDbConnection>().Use(c => new SqlConnection(configuration.SqlConnectionString));
+                configure.For<IDbConnection>().Use(c => ManagedIdentitySqlConnection.GetSqlConnection(configuration.SqlConnectionString, _azureServiceTokenProvider));
```
NickGraham101, line 46: Same issue raised in [this comment](https://github.com/SkillsFundingAgency/das-assessor-service/pull/1057/files#r496557271) and [this comment](https://github.com/SkillsFundingAgency/das-assessor-service/pull/1057/files#r517247524)

src/SFA.DAS.AssessorService.ExternalApiDataSync/Helpers/ManagedIdentitySqlConnection.cs
smneal: The ExteranApiDataSync has been moved to the assessor functions repo so is not part of this PR.

src/SFA.DAS.AssessorService.Application.Api/StartupConfiguration/Startup.cs
smneal: @adam230594 I don't think at this point we have actually generated a bearer token. It's only when we call the GetAccessTokenAsync in the ManagedIdentitySqlConnection class that the bearer token is generated.

src/SFA.DAS.AssessorService.Application.Api/StartupConfiguration/Startup.cs
smneal: It looks like ManagedIdentitySqlConnection.GetSqlConnection is called per request, so GetAccessTokenAsync would be called at the same time.

src/SFA.DAS.AssessorService.ExternalApiDataSync/Startup/Bootstrapper.cs
smneal: ExternalApiDataSync has been moved to the Assessor functions repo

src/SFA.DAS.AssessorService.ExternalApiDataSync/Startup/Bootstrapper.cs
smneal: ExternalApiDataSync has been moved to the Assessor functions repo

PR #1058, opened 2020-09-29
CON-2418 Adds additional validation of completion status to the api

ben1stone:
I have added two new validation rules to the CreateBatchCertificate request validator that will check if the learners completion status is not withdrawn or temporarily withdrawn

### DIFF
--- a/src/SFA.DAS.AssessorService.Application.Api/Validators/ExternalApi/Certificates/CreateBatchCertificateRequestValidator.cs
+++ b/src/SFA.DAS.AssessorService.Application.Api/Validators/ExternalApi/Certificates/CreateBatchCertificateRequestValidator.cs
@@ -58,11 +58,11 @@ namespace SFA.DAS.AssessorService.Application.Api.Validators.ExternalApi.Certifi
                     {
                         if (learnerDetails.CompletionStatus == (int)CompletionStatus.Withdrawn)
                         {
-                            context.AddFailure(new ValidationFailure("CompletionStatus", "Completion status must not be withdrawn"));
+                            context.AddFailure(new ValidationFailure("LearnerDetails", "Cannot find the apprentice details"));
                         }
                         else if (learnerDetails.CompletionStatus == (int)CompletionStatus.TemporarilyWithdrawn)
                         {
-                            context.AddFailure(new ValidationFailure("CompletionStatus", "Completion status must not be temporarily withdrawn"));
+                            context.AddFailure(new ValidationFailure("LearnerDetails", "Cannot find the apprentice details"));
                         }
                     }
                     else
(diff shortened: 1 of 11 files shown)


src/SFA.DAS.AssessorService.Application.Api/Validators/ExternalApi/Certificates/CreateBatchCertificateRequestValidator.cs
```diff
@@ -27,6 +27,7 @@ public class CreateBatchCertificateRequestValidator : AbstractValidator<BatchCer
                 {
                     var existingCertificate = await certificateRepository.GetCertificate(m.Uln, m.StandardCode);
                     var sumbittingEpao = await organisationQueryRepository.GetByUkPrn(m.UkPrn);
+                    var learnerDetails = await ilrRepository.Get(m.Uln, m.StandardId.GetValueOrDefault());
```
ChrisJWoodcock, line 30: If there is a possibility that the GetValueOrDefault() would return default null? then it is also possible the call to get the learner details would get nothing and you should handle that gracefully below

src/SFA.DAS.AssessorService.Domain/Consts/CompletionStatus.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace SFA.DAS.AssessorService.Domain.Consts
+{
+    public enum CompletionStatus
+    {
+        Continuing = 1,
+        Complete = 2,
+        Withdrawn = 3,
```
ChrisJWoodcock, line 11: Are these values being used as 'Magic numbers' elsewhere - could you correct that?

ChrisJWoodcock (changes requested):

src/SFA.DAS.AssessorService.Application/Handlers/Staff/GetLearnerDetailHandler.cs
```diff
@@ -119,16 +119,16 @@ private string FormatCompletionStatusDescription(int? completionStatus)
         {
             switch (completionStatus)
             {
-                case 1:
+                case (int)CompletionStatus.Continuing:
```
ChrisJWoodcock, line 122: This is definitely a use of it - but I was thinking about the original bug that you cannot use a withdrawn from the UI but could from the API - so wouldn't there be some logic which prevent this in the UI - and is that using magic numbers?

ChrisJWoodcock (approved):

src/SFA.DAS.AssessorService.Application.Api/Validators/ExternalApi/Epas/BatchEpaRequestValidator.cs
```diff
@@ -75,6 +83,8 @@ public BatchEpaRequestValidator(IStringLocalizer<BatchEpaRequestValidator> local
                                .Must(epa => outcomes.Contains(epa.EpaOutcome, StringComparer.InvariantCultureIgnoreCase)).WithMessage($"Invalid outcome: must be Pass, Fail or Withdrawn");
                     });
             });
+
+            //RuleFor(m => m.)
```
ChrisJWoodcock, line 87: Remove this

ChrisJWoodcock (changes requested):

ChrisJWoodcock (approved):

ChrisJWoodcock (approved):

PR #1063, opened 2020-10-06
CON-1094 Added a IlrsImport table, which is used for the LearnerDetailsImport (Schema Only)

ChrisJWoodcock:


src/SFA.DAS.AssessorService.Database/Tables/IlrsImport.sql
```diff
@@ -0,0 +1,38 @@
+﻿CREATE TABLE [dbo].[IlrsImport]
+(
+	[Id] UNIQUEIDENTIFIER NOT NULL DEFAULT NEWID() PRIMARY KEY,
```
cofaulco, line 3: bit of dodgy indentation on this file

cofaulco (approved):

cofaulco (approved):

PR #1068, opened 2020-10-21
Con 2263: Updates for ScheduleRun LastRunStatus 

hkf-tech:


src/SFA.DAS.AssessorService.Database/Tables/ScheduleRuns.sql
```diff
@@ -5,5 +5,6 @@
     [IsComplete] BIT NOT NULL DEFAULT 0, 
     [Interval] BIGINT NULL, 
     [IsRecurring] BIT NOT NULL DEFAULT 0, 
-    [ScheduleType] INT NOT NULL
+    [ScheduleType] INT NOT NULL, 
+    [LastRunStatus] INT NULL
```
cofaulco, line 9: is the intention to populate existing schedule runs and subsequently make `LastRunStatus` NOT NULL?

src/SFA.DAS.AssessorService.Data/ScheduleRepository.cs
```diff
@@ -47,25 +50,33 @@ public async Task<ScheduleRun> GetNextScheduledRun(int scheduleType)
                                     FROM ScheduleRuns 
                                     WHERE ScheduleType = @scheduleType 
                                     AND IsComplete = 0 
-                                    ORDER BY RunTime", new {scheduleType});
+                                    AND (LastRunStatus <> @scheduleRunStatus or LastRunStatus is null)
```
cofaulco, line 53: Should this filter out all except `Completed` runs?

ChrisJWoodcock (changes requested):

src/SFA.DAS.AssessorService.Data/ScheduleRepository.cs
hkf-tech: It is already filtering for Completed runs by IsComplete = 0, this is to stop getting the failed records

src/SFA.DAS.AssessorService.Database/Tables/ScheduleRuns.sql
hkf-tech: The intention is to allow null in LastRunStatus - As the newly created Schedule will have null. (Don't want to show any status  on UI if the job hasn't exectued yet).

src/SFA.DAS.AssessorService.Application.Api/Controllers/ScheduleController.cs
```diff
@@ -72,5 +73,12 @@ public async Task<IActionResult> DeleteScheduleRun(Guid scheduleRunId)
             await _scheduleRepository.DeleteScheduleRun(scheduleRunId);
             return Ok();
         }
+
+        [HttpPost("api/v1/schedule/updatelaststatus", Name = "UpdateLastRunStatus")]
+        public async Task<IActionResult> UpdateLastRunStatus([FromBody] UpdateLastRunStatusRequest request)
+        {
+            await _scheduleRepository.UpdateLastRunStatus(request);
```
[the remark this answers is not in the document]
hkf-tech: I just followed what the other action methods were doing - they are going directly to the repository atleast in the ScheduleController.
 I do agree it is better to go via handler - but i thnk it should be done for all the methods in the controller not for a single action method - a seperate tech debt should be raised for it.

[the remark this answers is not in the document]
ChrisJWoodcock: Ok - I see what you mean - do raise the tech debt though

ChrisJWoodcock (approved):

PR #1071, opened 2020-10-22
CON-2470 Changes

Howesy77:


ChrisJWoodcock (approved):
Approving this as the code does meet the AC's as they have **currently** been specified in the ticket regarding the privileges.

PR #1074, opened 2020-10-23
Con 2476 epao withdrawal   'which standard do you want to withdraw from assessing ' page

Howesy77:


src/SFA.DAS.AssessorService.Data/StandardRepository.cs
```diff
@@ -248,15 +248,16 @@ public async Task<int> GetEpaoStandardsCount(string endPointAssessorOrganisation
...
             {
                 PageOfResults = new List<EPORegisteredStandards>(),
                 TotalCount = 0
             };
             var total = await GetEpaoStandardsCount(endPointAssessorOrganisationId);
-            var skip = ((pageIndex ?? 1) - 1) * pageSize;
+            var skip = ((pageIndex ?? 1) - 1) * (pageSize ?? 10);
             var result = await _unitOfWork.Connection.QueryAsync<EPORegisteredStandards>(
                 "EPAO_Registered_Standards", 
                 param: new
```
cofaulco, line 263: Need to apply the same pagesize logic to the `Take` param. Also, can we just apply the "Default to 10" here rather than checking for null throughout the code?

cofaulco (commented):
should page size and page index have a default value or should the default be that you get 1 page of all standards unless a page size and index is specified?

src/SFA.DAS.AssessorService.Web/Views/ApplyForWithdrawal/ChooseStandardForWithdrawal.cshtml
```diff
@@ -42,16 +44,19 @@
...
+                                               sfa-validationerror-class="govuk-input--error"
+                                               asp-for="SelectedStandardForWithdrawal"
+                                               type="radio"
+                                               value="@item.StandardCode">
                                         <label class="govuk-label govuk-radios__label" asp-for="SelectedStandardForWithdrawal">
-                                            @item.StandardName
+                                            @item.StandardName (Level @item.Level)
                                         </label>
+                                        <span class="govuk-hint govuk-radios__hint">
+                                            @($"STO{@item.StandardCode}")
```
ChrisJWoodcock, line 58: I think this is the reference number here? which is prefixed already

ChrisJWoodcock (changes requested):

ChrisJWoodcock (approved):

cofaulco (approved):

PR #1075, opened 2020-10-26
Con 2596 alert email

VasanthaKasirajan3008:


src/SFA.DAS.AssessorService.Application/Handlers/Apply/SubmitApplicationSequenceHandler.cs
```diff
@@ -206,9 +206,6 @@ private async Task NotifyContact(Domain.Entities.Contact contactToNotify, ApplyD
             {
                 var emailTemplate = await _eMailTemplateQueryRepository.GetEmailTemplate(EmailTemplateNames.ApplyEPAOInitialSubmission);
                 await _mediator.Send(new SendEmailRequest(email, emailTemplate, new { contactname, reference }), cancellationToken);
-
-                var emailTemplateAlert = await _eMailTemplateQueryRepository.GetEmailTemplate(EmailTemplateNames.ApplyEPAOAlertSubmission);
```
ChrisJWoodcock, line 210: Have you just removed this code? - why?

ChrisJWoodcock (changes requested):

src/SFA.DAS.AssessorService.Application/Handlers/Apply/SubmitApplicationSequenceHandler.cs
VasanthaKasirajan3008: have introduced this  code as part of Tech Debt - Email Template Recipients . As part of the work we extracted recipients  to the new table. There was a scenario where the bug been raised and code been introduced as part of the fix.

It was confused as that point  why to write code because all we did was database changes.

cofaulco: I think this code is still needed. The way I am reading the code is that if we supply an email address in the request then it sends that email to that user. Otherwise it will send it to the recipients held in the DB. I think, in this scenario we do not need to pass the emails in the Alert SendEmailRequest rather than removing the code. What do you think @ChrisJWoodcock ??

ChrisJWoodcock: @cofaulco exactly that - @VasanthaKasirajan3008 - I have documented that on a comment on the Jira ticket

cofaulco (approved):

ChrisJWoodcock (approved):

PR #1076, opened 2020-10-29
Update Sequence.cshtml

Howesy77:


ChrisJWoodcock (changes requested):
I don't see a code change for the email in scenario 3 and the context text 'Withdrawing from register' - in  scenario 1

ChrisJWoodcock (approved):

PR #1077, opened 2020-11-06
CON-2264 Move print batch create to Assessor

ChrisJWoodcock:


src/SFA.DAS.AssessorService.Application.UnitTests/Handlers/Certificates/Query/WhenGetCertificatesToBePrinted.cs
```diff
@@ -51,6 +51,6 @@ public void then_certificates_are_returned()
         public void then_no_failed_certificates_are_returned()
         {
             _result.Certificates.Select(r => r.OverallGrade).Should().NotContain(CertificateGrade.Fail);
-        }
+        }*/
```
cofaulco, line 54: Commented code?

src/SFA.DAS.AssessorService.Application/Interfaces/IBatchLogRepository.cs
```diff
@@ -10,6 +10,8 @@ namespace SFA.DAS.AssessorService.Application.Interfaces
     public interface IBatchLogRepository
     {
         Task<BatchLog> Create(BatchLog batchLog);
-        Task<ValidationResponse> UpdateBatchLogBatchWithDataRequest(Guid requestId, BatchData requestBatchData);
+        Task<ValidationResponse> UpdateBatchLogSentToPrinter(int batchNumber, DateTime batchCreated, int numberOfCertificates,
```
cofaulco, line 13: could we pass a BatchData object of some kind here rather than loads of separate arguments?

src/SFA.DAS.AssessorService.Application.UnitTests/Handlers/Certificates/Query/WhenGetCertificatesToBePrinted.cs
ChrisJWoodcock: Added additional tests

src/SFA.DAS.AssessorService.Application/Interfaces/IBatchLogRepository.cs
ChrisJWoodcock: Done

cofaulco (approved):

cofaulco (approved):

cofaulco (approved):

cofaulco (approved):

PR #1084, opened 2020-11-25
CON-1561 Fixes query for removed column; disable data fix for initial database

ChrisJWoodcock:
…e setup.


src/SFA.DAS.AssessorService.Database/PreDeploymentScripts/con-1561_update_staff_reports_id.sql
```diff
@@ -17,7 +17,9 @@ DECLARE @DefaultEPAORegister UNIQUEIDENTIFIER = '4D07FE96-AE3E-4476-9A53-673E193
 DECLARE @DefaultRegisterListOfOrganisations UNIQUEIDENTIFIER = '3E230675-D61D-4EF0-A678-A254F77C58B7'
 DECLARE @DefaultRegisterListOfStandards UNIQUEIDENTIFIER = '37770D79-1733-43F9-90D5-88F0609DD7E2'
 
-IF((SELECT COUNT(*) FROM [StaffReports] WHERE Id NOT IN
+IF(EXISTS(SELECT * FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_NAME = 'EMailTemplates'))
```
cofaulco, line 20: Is this supposed to be `TABLE_NAME = 'StaffReports'`?

cofaulco (changes requested):

src/SFA.DAS.AssessorService.Database/PreDeploymentScripts/con-1561_update_staff_reports_id.sql
ChrisJWoodcock: No... Fixed

cofaulco (approved):

PR #1087, opened 2020-12-01
FAT2-287-epao-register

scottywakefield:
The query in question was not populating all values in the poco, `AssessmentOrganisationSummary` hence null was returned when calling the api. This has now been fixed.


src/SFA.DAS.AssessorService.Data/RegisterQueryRepository.cs
```diff
@@ -126,7 +126,18 @@ public async Task<IEnumerable<AssessmentOrganisationSummary>> GetAssessmentOrgan
...
+                        "EndPointAssessorUkprn as ukprn, " +
+                        "OrganisationData, " +
+                        "OrganisationTypeId, " +
+                        "ot.Type as OrganisationType, " +
+                        "o.Status, " +
+                        "JSON_VALUE(OrganisationData,'$.Email') as Email " +
+                        "from [Organisations] o " +
+                        "join OrganisationType ot " +
+                        "on ot.id = o.OrganisationTypeId " +
+                        "and ot.Status = 'Live'");
```
dashton82, line 140: Just a check but is this similar to the behaviour of others? should we not return all and filter after on the OrganisationType status

scottywakefield: I can remove that, but it's making sure that if there are 2 org types, you only use the org type that is live. It's not the status of the org, but rather of the org type.

dashton82 (approved):

PR #1090, opened 2020-12-07
CON-2595 Update will org be completing assessments page

ChrisJWoodcock:
… completing assesments page


src/SFA.DAS.AssessorService.Web/Controllers/Apply/ApplicationService.cs
```diff
@@ -69,6 +118,8 @@ public async Task<CreateApplicationRequest> BuildCreateApplicationRequest(string
 
     public interface IApplicationService
```
cofaulco, line 119: Can we split the interface out into a separate file to be consistent with everywhere else?

cofaulco (commented):
There don't appear to be any new/updated tests for this work as I would expect to see considering the changes.

src/SFA.DAS.AssessorService.Web/Controllers/Apply/ApplicationService.cs
ChrisJWoodcock: Done

ChrisJWoodcock:
> There don't appear to be any new/updated tests for this work as I would expect to see considering the changes.

I've added unit test for the pipelines count and starting a withdrawal application

cofaulco (approved):

PR #1091, opened 2020-12-09
Con 2610 epao assessor confirmation

kola-samuel:


src/SFA.DAS.AssessorService.Application/Handlers/Apply/CreateApplicationHandler.cs
```diff
@@ -68,7 +68,7 @@ public async Task<Guid> Handle(CreateApplicationRequest request, CancellationTok
 
             return Guid.Empty;
         }
-
+
```
ChrisJWoodcock, line 71: Remove this - blank lines are ok but not spaces

src/SFA.DAS.AssessorService.ApplyTypes/Application.cs
```diff
@@ -130,4 +130,5 @@ public class Submission
         public Guid SubmittedBy { get; set; }
         public string SubmittedByEmail { get; set; }
     }
+
```
ChrisJWoodcock, line 133: Unnecessary change

src/SFA.DAS.AssessorService.Web/Controllers/Apply/ApplicationController.cs
```diff
@@ -734,6 +739,21 @@ public async Task<IActionResult> Feedback(Guid id)
             return View("~/Views/Application/Feedback.cshtml", sequenceVm);
         }
 
+        private ApplicationSubmissionType GetApplicationSubmissionTypeAsync(ApplicationResponse application, Sequence sequence)
```
ChrisJWoodcock, line 742: There is already an ApplicationTypes class which has definitions for types of application StandardWithdrawal etc - so we should not have something new for that.

Other views e.g. 'Feedback.cshtml' are just using the ApplyConst directly - so that would be more consistent if you put the SequenceNo in the view model and the test against the ApplyConst which is also used in dozens of places, so it needs to kept consistent

src/SFA.DAS.AssessorService.Web/Views/Application/Submitted.cshtml
```diff
@@ -1,56 +1,139 @@
...
+
+
+                    <h2 class="govuk-heading-m govuk-!-margin-top-6">
+                        What happens next
+                    </h2>
+
+
+                    @if (Model.ApplicationSubmissionType == ApplicationSubmissionType.WithdrawalFromRegister)
+                    {
+                        <p class="govuk-body">
```
ChrisJWoodcock, line 41: This part looks the same below so should not be duplicated

src/SFA.DAS.AssessorService.Web/Controllers/Apply/ApplicationController.cs
```diff
@@ -25,6 +25,7 @@
 using SFA.DAS.AssessorService.Web.ViewModels.Apply;
 using SFA.DAS.QnA.Api.Types;
 using SFA.DAS.QnA.Api.Types.Page;
+using AssessorServiceApplyType = SFA.DAS.AssessorService.ApplyTypes;
```
ChrisJWoodcock, line 28: Is this used?

ChrisJWoodcock (changes requested):

src/SFA.DAS.AssessorService.Application/Handlers/Apply/CreateApplicationHandler.cs
kola-samuel: removed

src/SFA.DAS.AssessorService.Web/Controllers/Apply/ApplicationController.cs
kola-samuel: updated

src/SFA.DAS.AssessorService.Web/Views/Application/Submitted.cshtml
kola-samuel: updated

src/SFA.DAS.AssessorService.Web/Controllers/Apply/ApplicationController.cs
kola-samuel: removed

ChrisJWoodcock (approved):

PR #1093, opened 2020-12-11
CON-2601 When do you want to withdraw page changes

ChrisJWoodcock:


src/SFA.DAS.AssessorService.Application/Handlers/OrganisationHandlers/GetEarliestWithdrawalDateHandler.cs
```diff
@@ -0,0 +1,25 @@
...
+        
+        public GetEarliestWithdrawalDateHandler(IDateTimeHelper dateTimeHelper)
+        {
+            _dateTimeHelper = dateTimeHelper;
+        }
+
+        public async Task<DateTime> Handle(GetEarliestWithdrawalDateRequest request, CancellationToken cancellationToken)
+        {
+            // the earliest withdrawal is currently 6 months from the current date for any organisation or standard
+            return await Task.FromResult(_dateTimeHelper.DateTimeNow.AddMonths(6));
```
cofaulco, line 22: should we make this a configuration value rather than hardcode to 6 months

cofaulco (approved):

src/SFA.DAS.AssessorService.Application.Api/Controllers/OrganisationQueryController.cs
```diff
@@ -137,5 +137,23 @@ public async Task<IActionResult> GetOrganisationForContact(Guid userId)
...
+        [SwaggerResponse((int)HttpStatusCode.InternalServerError, Type = typeof(ApiResponse))]
+        public async Task<IActionResult> GetOrganisationEarliestWithdrawal(Guid id, int? standardId = null)
+        {
+            if (standardId.HasValue)
+            {
+                _logger.LogInformation($"Received request to retrieve earliest withdrawal for Standard: {standardId} of Organisation: {id}");
+            }
+            else
+            {
+                _logger.LogInformation($"Received request to retrieve earliest withdrawal for Organisation: {id}");
```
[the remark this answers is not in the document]
ChrisJWoodcock: This would be a lot of rework and the pipelines count is implemented like this as are several other endpoints - so I think it's ok to leave as is

src/SFA.DAS.AssessorService.Application/Handlers/OrganisationHandlers/GetEarliestWithdrawalDateHandler.cs
ChrisJWoodcock: If it does change I think it may change more than just adjusting the months to be more specific to an organisation or a standard or other logic - so ok to leave as is for now.

PR #1096, opened 2020-12-15
CON-2600 how will you communicate your exit

ben1stone:


ChrisJWoodcock (approved):
This will be manually regression tested to see if anything bizarre occurs in the organisation journey as the design of this page is exploratory

PR #1100, opened 2020-12-17
CON-2790 CON-2805 Updated application response to contain application type directly

ChrisJWoodcock:


src/SFA.DAS.AssessorService.Api.Types/Models/Apply/ApplicationResponse.cs
```diff
@@ -21,44 +22,10 @@ public class ApplicationResponse
...
-        public bool IsSequenceActive(int sequenceNo)
-        {
-            // a sequence can be considered active even if it does not exist in the ApplyData, since it has not yet been submitted and is in progress.
-            return ApplyData?.Sequences?.Any(x => x.SequenceNo == sequenceNo && x.IsActive) ?? true;
-        }
-
         public bool IsWithdrawalApplication => IsStandardWithdrawalApplication || IsOrganisationWithdrawalApplication;
 
-        public bool IsStandardWithdrawalApplication => GetApplicationType() == ApplicationTypes.StandardWithdrawal;
+        public bool IsStandardWithdrawalApplication => ApplicationType == ApplicationTypes.StandardWithdrawal;
```
kola-samuel, line 27: We should also have  IsStandardApplication and  IsSOrganizationApplication fields

kola-samuel (changes requested):
More Application type Fields required

src/SFA.DAS.AssessorService.Api.Types/Models/Apply/ApplicationResponse.cs
ChrisJWoodcock: There are properties for it - I hadn't added them here yet as I didn't have a use - there isn't an organisation application as it cant exists, there is a combined application [(organisation + financial + standard), (organisation + standard), (financial + standard)] or a standard application

ChrisJWoodcock:
> More Application type Fields required

Covered that in my comment above

cofaulco (approved):

kola-samuel (approved):

PR #1104, opened 2020-12-22
CON-2832 Prevent serialization of cyclic dependency between Certificate and CertificateBatchLog

Howesy77:


cofaulco (approved):

src/SFA.DAS.AssessorService.Domain/Entities/Certificate.cs
```diff
@@ -36,7 +36,9 @@ public class Certificate : BaseEntity
         public bool IsPrivatelyFunded { get; set; }
         public string PrivatelyFundedStatus { get; set; }
         public DateTime CreateDay { get; set; }
+        [JsonIgnore]
```
ChrisJWoodcock, line 39: Will this just not serialize the collection of CertificateLogs? I expect there are cases where they are used after having been serialized and then deserialized - have you accounted for that?

ChrisJWoodcock (changes requested):

src/SFA.DAS.AssessorService.Domain/Entities/Certificate.cs
Howesy77: The certificate logs object has cyclic references so fails to get serialized. TBH I'm not massively impressed with the method o serializing and reserializing an object in the same method?

ChrisJWoodcock (approved):

PR #1106, opened 2021-01-06
Con 2935 feedback notification email

Howesy77:


src/SFA.DAS.AssessorService.Application/Handlers/Apply/Review/ReturnApplicationSequenceHandler.cs
```diff
@@ -104,13 +104,39 @@ private async Task NotifyContact(Guid applicationId, int sequenceNo, Cancellatio
                 }
                 else if(sequenceNo == ApplyConst.ORGANISATION_WITHDRAWAL_SEQUENCE_NO)
                 {
-                    // TO DO: Send an email to notify organisation withdrawal approval to EPAO
                     var lastSubmission = application.ApplyData?.Apply.LatestOrganisationWithdrawalSubmission;
```
ChrisJWoodcock, line 107: The code for both cases is almost the same - these could be combined with just the last submission part extracted

src/SFA.DAS.AssessorService.Application/Handlers/Apply/Review/ReturnApplicationSequenceHandler.cs
```diff
@@ -104,13 +104,39 @@ private async Task NotifyContact(Guid applicationId, int sequenceNo, Cancellatio
...
+                    if (lastSubmission != null)
```
ChrisJWoodcock, line 108: We should only be sending the email in the case of this being feedback and not for approval - there is switch above that check for that - similar can be used here

ChrisJWoodcock (changes requested):

src/SFA.DAS.AssessorService.Application/Handlers/Apply/Review/ReturnApplicationSequenceHandler.cs
Howesy77: Updated code

src/SFA.DAS.AssessorService.Application/Handlers/Apply/Review/ReturnApplicationSequenceHandler.cs
Howesy77: Done

ChrisJWoodcock (approved):

PR #1113, opened 2021-01-28
FAT2-371 remove fatv1 api

dashton82:
This removes FATv1 from the assessor service. Due to large amounts of change occurring as a result of the work the Standards Versioning (SV) team are doing. Only the call for Standards Collation has been modified to use the assessor outer api. This will eventually change further as the SV project progresses. The call that was being done to check a provider name has also been replaced to call RoATP directly for now. This will ultimately be replaced with the work SV team are doing.


dashton82:
The sonar fail is due to me touching an existing set of handlers that were already pretty much a CTRL C, CTRL V from one an other.....

src/SFA.DAS.AssessorService.Application/Infrastructure/OuterApi/ApiClient.cs
```diff
@@ -0,0 +1,56 @@
...
+        private void AddHeaders()
+        {
+            _httpClient.DefaultRequestHeaders.Add("Ocp-Apim-Subscription-Key", _config.Key);
+            _httpClient.DefaultRequestHeaders.Add("X-Version", "1");
+        }
+    }
+
+    public interface IApiClient
+    {
+        Task<TResponse> Get<TResponse>(IGetApiRequest getStandardsRequest);
```
ChrisJWoodcock, line 54: The name should be request

ChrisJWoodcock (changes requested):

ChrisJWoodcock (dismissed):

scottywakefield (approved):

src/SFA.DAS.AssessorService.Application/Infrastructure/OuterApi/ApiClient.cs
dashton82: all resolved now

ChrisJWoodcock (approved):

PR #1116, opened 2021-02-26
CON-3139 Correctly identify initial applications distinct from financial health checks

ChrisJWoodcock:
…h check standard applications.


src/SFA.DAS.AssessorService.Application/Mapping/CustomResolvers/ApplicationTypeResolver.cs
```diff
@@ -11,13 +11,26 @@ public class ApplicationTypeResolver : IValueResolver<ApplySummary, ApplicationR
     {
         public string Resolve(ApplySummary source, ApplicationResponse destination, string destMember, ResolutionContext context)
         {
-            if (IsSequenceRequired(source.ApplyData, ApplyConst.ORGANISATION_SEQUENCE_NO) && IsSequenceRequired(source.ApplyData, ApplyConst.STANDARD_SEQUENCE_NO))
+            bool OrganisationSectionRequired(int sectionNo) =>
```
cofaulco, line 14: I think that this logic needs to be covered by some unit tests

cofaulco (changes requested):

src/SFA.DAS.AssessorService.Application/Mapping/CustomResolvers/ApplicationTypeResolver.cs
ChrisJWoodcock: Done

cofaulco (approved):

PR #1119, opened 2021-03-05
SV-488 get standard options

ben1stone:


src/SFA.DAS.AssessorService.Application.Api.External/Models/Response/Standards/StandardOptions.cs
```diff
@@ -6,6 +6,7 @@ namespace SFA.DAS.AssessorService.Application.Api.External.Models.Response.Stand
 {
     public sealed class StandardOptions : IEquatable<StandardOptions>
     {
+        public string StandardUId { get; set; }
```
DevenKShah, line 9: @alandburns We have included the `StandardUId` in outer response, but should we be introducing a version 2 with updated fields names, StandardCode => LarsCode and StandardReference => IFateReferenceNumber? Also this will feed into how the options by standard endpoint works.

DevenKShah (dismissed):
Looks good

DevenKShah (approved):

PR #1120, opened 2021-03-05
SV-487 Import standards data into new table

DevenKShah:
~~This PR has breaking changes, where importing of standards in old ways is no longer working.~~

The PR now updates the existing tables as well so as to not break any existing functionality.

I have deleted the External API project that was mainly created for IFATE call. Couple of files `CacheService` and `EntityNotFoundException` where in this project and still have valid reference. Hence I have moved them in to appropriate projects. 

I have reused `update-standards` endpoint so that the time triggered functions carry on working without any change. 

The original import was checking for a minimum time interval of 23 hours before we can run the import. I have removed this as to allow us to run the function ad-hoc if necessary. 

The import strategy is left to be simple, delete-all insert-all for now. This we can revisit if need be. But the idea is that any safety handles that are deemed required should be implemented in the courses api.


CodeWolf86:
Just to clarify.
- We've removed the IFA External Client code which we intended to do
- We've also removed the existing Collation Standards Import as a result
- The new standards table will be populated by the same function as before

So while this new release would begin the population of the new data structure ( only the standard table has been created out of the ones that have been designed thus far ) We are breaking the connection of the existing import. So the existing standard collation tables that are in use will become stale over time.

So methods such as UpsertApprovedStandards which were originally used have been left alone even though they are no longer called by anything outside of unit tests.

As a result is the plan to deploy shortly after such that the stale data has no impact ( aware it broke for 2 weeks and no one noticed for example ) and just live with it.

Or is it to do a secondary update such that we update the existing collation tables but via our new outer api as an interim? essentially doing a dual import? this would allow a much more gradual roll out as the database structures could live side by side.

CodeWolf86 (changes requested):
Just to clarify.
- We've removed the IFA External Client code which we intended to do
- We've also removed the existing Collation Standards Import as a result
- The new standards table will be populated by the same function as before

So while this new release would begin the population of the new data structure ( only the standard table has been created out of the ones that have been designed thus far ) We are breaking the connection of the existing import. So the existing standard collation tables that are in use will become stale over time.

So methods such as UpsertApprovedStandards which were originally used have been left alone even though they are no longer called by anything outside of unit tests.

As a result is the plan to deploy shortly after such that the stale data has no impact ( aware it broke for 2 weeks and no one noticed for example ) and just live with it.

Or is it to do a secondary update such that we update the existing collation tables but via our new outer api as an interim? essentially doing a dual import? this would allow a much more gradual roll out as the database structures could live side by side.

DevenKShah:
Before the plan was that this will be a big release where we break everything and rebuild all at one go. This was my attempt to keep PR's smaller and manageable. However, post the PR we came to

src/SFA.DAS.AssessorService.Application.Api/Services/OuterApiService.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            var response = await outerApiClient.Get<GetStandardsListResponse>(new GetActiveStandardsRequest());
+            return response.Standards;
+        }
+
+        public async Task<IEnumerable<GetStandardByIdResponse>> GetAllStandardDetails(IEnumerable<string> standardUIds)
+        {
+            try
+            {
+                var tasks = standardUIds.Select(id => outerApiClient.Get<GetStandardByIdResponse>(new GetStandardByIdRequest(id))).ToList();
```
CodeWolf86, line 32: As mentioned previously I get errors sometimes when running this locally and in Visual Studio. Running from command line seems to mitigate the issue somewhat. Will be interested to see what this is like when in an environment, specifically as to whether connection pooling is required in the course api to deal with the volume of requests.

CodeWolf86 (approved):
Approved, Just a couple of questions which may need a tweak otherwise really good work on the import @DevenKShah  Know it's been a trial on this branch but it's a nice rework which will allow us to simplify later down the line. 🚀

src/SFA.DAS.AssessorService.Data/StandardRepository.cs
```diff
@@ -154,7 +187,7 @@ private async Task<List<StandardNonApprovedCollation>> GetStandardNonApprovedCol
 
         public async Task<DateTime?> GetDateOfLastStandardCollation()
         {
-            const string sql = "SELECT TOP 1 COALESCE(MAX(DateUpdated), MAX(DateAdded)) MaxDate FROM [StandardCollation]";
+            const string sql = "SELECT TOP 1 COALESCE(MAX(DateUpdated), MAX(DateAdded)) MaxDate FROM [Standard]";
```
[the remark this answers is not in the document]
DevenKShah: That's a good catch. I removed the check of 23 hours before the data load can be triggered again. As a result this method was made redundant. I have deleted it.

src/SFA.DAS.AssessorService.Application.Api/Services/OuterApiService.cs
DevenKShah: agree

PR #1123, opened 2021-03-15
SV-536 Add Ability to Get Options for Standard By larsCode / IFate Reference Number

CodeWolf86:


src/SFA.DAS.AssessorService.Application/Infrastructure/OuterApi/GetStandardByIdResponse.cs
```diff
@@ -0,0 +1,168 @@
...
+        public string IntegratedDegree { get; set; }
+
+        [JsonProperty("sectorSubjectAreaTier2")]
+        public decimal SectorSubjectAreaTier2 { get; set; }
+
+        [JsonProperty("sectorSubjectAreaTier2Description")]
+        public string SectorSubjectAreaTier2Description { get; set; }
+
+        [JsonProperty("apprenticeshipFunding")]
+        public List<ApprenticeshipFunding> ApprenticeshipFunding { get; set; }
```
DevenKShah, line 70: Don't think we need this information in assessor service.

DevenKShah (approved):
Looks good. I am just questioning if we should only restrict to deserialize the details that we are actually consuming.

CodeWolf86:
> Looks good. I am just questioning if we should only restrict to deserialize the details that we are actually consuming.

As the options information came from the GetById Standard Endpoint and it returned the detailed response I put in everything that came back in case it was required further down the line. ( Ie where we now query by ID to get additional information rather than relying on the cache. ) 

It's worth noting that the GetOptions endpoint on the external Api, that though you can provide it with an IFate Ref or a LarsCode, if you supply a StandardUID it also still works as it's a string and just gets passed straight through given the nature of the underlying Apis. ( The current swagger documentation on the endpoint only states it accepts the first two so made a decision to not update that yet )

As to whether it's worth cutting down the information deserialised, I can cut it back to what exists on the Get All with a few extras required for your looped caching if we want a middle ground until the data is required?

DevenKShah:
Including all the key identifier and VersionDetail fields, these are what we need. May be we can limit it to this? 
```
public class StandardData
    {
        public int? Level { get; set; }
        public string Category { get; set; }
        public string IfaStatus { get; set; }

        public DateTime? EffectiveFrom { get; set; }
        public DateTime? EffectiveTo { get; set; }
        public DateTime? LastDateForNewStarts { get; set; }
        public bool IfaOnly { get; set; }
        public string EqaProviderName { get; set; }
        public string EqaProviderContactName { get; set; }
        public string EqaProviderContactAddress { get; set; }
        public string EqaProviderContactEmail { get; set; }
        public string EqaProviderWebLink { get; set; }
        public string IntegratedDegree { get; set; }
        public int? Duration { get; set; }
        public int? MaxFunding { get; set; }
        public string Trailblazer { get; set; }
        public string Ssa1 { get; set; }
        public string Ssa2 { get; set; }
        public string OverviewOfRole { get; set; }
        public DateTime? PublishedDate { get; set; }
        public bool? IsPublished { get; set; }

        public bool? IsActiveStandardInWin { get; set; }

        public string FatUri { get; set; }
        public string IfaUri { get; set; }

        public string AssessmentPlanUrl { get; set; }
        public string StandardPageUrl { get; set; }
    }
```

PR #1127, opened 2021-03-22
SV-535 Fix version to be a string in external api response

DevenKShah:
The external API should match the version field to the IFATE API, hence I am converting it  in the internal API. Currently the format in IFATE API is matching to ("#.0"). If this ever changes our code will have to change as well.


CodeWolf86 (approved):
Approved on the basis this resolves the immediate issue, and returns a string to the external api.

It won't work for version numbers with 2 decimal digits, but there aren't any yet, ( we're at 1.3, so not quite near 1.10 )

With a tech debt ticket to be raised to store and expose version as a string and then rework how the course api orders on version

PR #1131, opened 2021-03-25
SV-537 get standard version options

ben1stone:
I've created the get standard version endpoint and added a unit test project for the external API.

To get the StandardUId I have found all standards with the given standard reference and then tried to match with the given version. I thought this was better than querying based on the standard reference and version as the original query could be reused or it might help if we need error handling that says "version x does not exist for that standard".

Once we know the StandardUId I could use the existing outer API method

### DIFF
--- a/src/SFA.DAS.AssessorService.Application.Api.External/Controllers/StandardsController.cs
+++ b/src/SFA.DAS.AssessorService.Application.Api.External/Controllers/StandardsController.cs
@@ -66,5 +66,27 @@ namespace SFA.DAS.AssessorService.Application.Api.External.Controllers
 
             return Ok(requestedStandard);
         }
+
+        [HttpGet("options/{standardReference}/{version}")]
+        [SwaggerResponseExample((int)HttpStatusCode.OK, typeof(SwaggerHelpers.Examples.GetStandardVersionOptionsResponseExample))]
+        [SwaggerResponse((int)HttpStatusCode.OK, "The list of options.", typeof(StandardOptions))]
+        [SwaggerResponse((int)HttpStatusCode.NoContent, "The standard version was found, however it has no options.")]
+        [SwaggerResponse((int)HttpStatusCode.NotFound, "The standard version was not found.")]
+        [SwaggerOperation("Get Options for a standard version", "Gets the latest list of course options for the specified Standard version.", Produces = new string[] { "application/json" })]
+        public async Task<IActionResult> GetOptionsForStandardVersion(string standardReference, decimal version)
+        {
+            var standardVersion = await _apiClient.GetStandardOptionsByStandardReferenceAndVersion(standardReference, version);
+
+            if (standardVersion is null)
+            {
+                return NotFound();
+            }
+            else if (standardVersion.CourseOption is null)
+            {
+                return NoContent();
+            }
+
+            return Ok(standardVersion);
+        }
     }
 }
--- a/src/SFA.DAS.AssessorService.Application.Api.External/Controllers/StandardsController.cs
+++ b/src/SFA.DAS.AssessorService.Application.Api.External/Controllers/StandardsController.cs
@@ -81,7 +81,7 @@ namespace SFA.DAS.AssessorService.Application.Api.External.Controllers
             {
                 return NotFound();
             }
-            else if (standardVersion.CourseOption is null)
+            else if (standardVersion.CourseOption is null || standardVersion.CourseOption.Any() == false)
             {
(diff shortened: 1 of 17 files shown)


src/SFA.DAS.AssessorService.Application.Api.External/Controllers/StandardsController.cs
```diff
@@ -66,5 +66,27 @@ public async Task<IActionResult> GetOptionsForStandard([SwaggerParameter("Standa
...
             return Ok(requestedStandard);
         }
+
+        [HttpGet("options/{standardReference}/{version}")]
+        [SwaggerResponseExample((int)HttpStatusCode.OK, typeof(SwaggerHelpers.Examples.GetStandardVersionOptionsResponseExample))]
+        [SwaggerResponse((int)HttpStatusCode.OK, "The list of options.", typeof(StandardOptions))]
+        [SwaggerResponse((int)HttpStatusCode.NoContent, "The standard version was found, however it has no options.")]
+        [SwaggerResponse((int)HttpStatusCode.NotFound, "The standard version was not found.")]
+        [SwaggerOperation("Get Options for a standard version", "Gets the latest list of course options for the specified Standard version.", Produces = new string[] { "application/json" })]
+        public async Task<IActionResult> GetOptionsForStandardVersion(string standardReference, decimal version)
```
CodeWolf86, line 76: Should version still be a string? otherwise ( down the line ) if they specify 1.10 in the query string, decimal makes that 1.1 which would be an older version?

src/SFA.DAS.AssessorService.Application.Api/Controllers/StandardServiceController.cs
```diff
@@ -69,5 +69,16 @@ public async Task<IActionResult> GetStandardOptionsForStandard(string id)
             var standardOptions = await _standardService.GetStandardOptionsByStandardId(id);
             return Ok(standardOptions);
         }
+
+        [HttpGet("standard-options/{standardReference}/{version}", Name = "GetStandardOptionsByStandardReferenceAndVersion")]
+        [SwaggerResponse((int)HttpStatusCode.OK, Type = typeof(StandardOptions))]
+        [SwaggerResponse((int)HttpStatusCode.BadRequest, Type = typeof(IDictionary<string, string>))]
+        [SwaggerResponse((int)HttpStatusCode.InternalServerError, Type = typeof(ApiResponse))]
+        public async Task<IActionResult> GetStandardOptionsForStandardReferenceAndVersion(string standardReference, decimal version)
```
CodeWolf86, line 77: Same as mentioned previously, the draft specification isn't explicit in mentioning decimal or string

CodeWolf86 (approved):

DevenKShah (changes requested):
Since there is no need for separate validation messages, I think it will be easier to query the standards table with refnum and version together and return not found if that comes back with no result. That way we avoid any extra logic in the code.

src/SFA.DAS.AssessorService.Application.Api.External/Controllers/StandardsController.cs
DevenKShah: Agree version should be a string

src/SFA.DAS.AssessorService.Application.Api.External/Infrastructure/ApiClient.cs
```diff
@@ -226,6 +226,12 @@ public virtual async Task<StandardOptions> GetStandardOptionsByStandard(string s
             return response;
         }
 
+        public virtual async Task<StandardOptions> GetStandardOptionsByStandardReferenceAndVersion(string standardReference, decimal version)
```
DevenKShah, line 229: Same here versions should be string

src/SFA.DAS.AssessorService.Application.Api.External/Controllers/StandardsController.cs
```diff
@@ -66,5 +66,27 @@ public async Task<IActionResult> GetOptionsForStandard([SwaggerParameter("Standa
 
             return Ok(requestedStandard);
         }
+
+        [HttpGet("options/{standardReference}/{version}")]
```
CodeWolf86, line 70: Just ran this locally, the url doesn't adhere to the other endpoints, should be /api/v1/standards/options/{standardref}/{version}

CodeWolf86 (changes requested):

CodeWolf86 (approved):

PR #1132, opened 2021-03-29
SV-487 add resilience to outer api

DevenKShah:


src/SFA.DAS.AssessorService.Application.Api/StartupConfiguration/Startup.cs
```diff
@@ -169,9 +171,12 @@ public IServiceProvider ConfigureServices(IServiceCollection services)
                     })
                     .SetHandlerLifetime(TimeSpan.FromMinutes(5));
 
-                
-                services.AddHttpClient<OuterApiClient>().SetHandlerLifetime(TimeSpan.FromMinutes(5));
-                
+                services
+                    .AddHttpClient<OuterApiClient>()
+                    .SetHandlerLifetime(TimeSpan.FromMinutes(5))
+                    .AddPolicyHandler(GetRetryPolicy())
```
CodeWolf86, line 177: If we do `.AddPolicyHandler` does the Retry Policy Wrap the Circuit Breaker Policy? To be certain could we have one `AddPolicyHandler(GetWrappedPolicy())` call and then that does the wrapping https://github.com/App-vNext/Polly/wiki/PolicyWrap

CodeWolf86 (changes requested):

CodeWolf86 (dismissed):

src/SFA.DAS.AssessorService.Application.Api/StartupConfiguration/Startup.cs
CodeWolf86: Changed to approve as AddPolicyHandler Wraps

PR #1133, opened 2021-03-29
SV-578 strip out privately funded logic and pages

CodeWolf86:


src/SFA.DAS.AssessorService.Application.Api.Client/Clients/CertificateApiClient.cs
```diff
@@ -52,8 +44,7 @@ public async Task<Certificate> UpdateCertificate(UpdateCertificateRequest certif
             }
         }
 
-        public async Task<CertificateAddress> GetContactPreviousAddress(string userName,
-            bool isPrivatelyFunded)
+        public async Task<CertificateAddress> GetContactPreviousAddress(string userName,bool isPrivatelyFunded)
         {
             using (var httpRequest = new HttpRequestMessage(HttpMethod.Get, $"api/v1/certificates/contact/previousaddress?username={WebUtility.UrlEncode(userName)}&isPrivatelyFunded={isPrivatelyFunded}"))
```
ChrisJWoodcock, line 49: Still using privately funded?

src/SFA.DAS.AssessorService.Application/Handlers/Search/SearchResultExtensions.cs
```diff
@@ -60,14 +60,14 @@ public static List<SearchResult> MatchUpExistingCompletedStandards(this List<Sea
...
                 var createdLogEntry = certificateLogs.FirstOrDefault(l => l.Status == CertificateStatus.Draft);
-                if (request.IsPrivatelyFunded)
+                var toBeApprovedLogEntry = certificateLogs.FirstOrDefault(l => l.Status == CertificateStatus.ToBeApproved);
+
+                if(toBeApprovedLogEntry != null)
                 {
-                    var toBeApprovedLogEntry = certificateLogs.FirstOrDefault(l => l.Status == CertificateStatus.ToBeApproved);
-                    if (toBeApprovedLogEntry == null)
-                        continue;
+                    // To Be Approved only exists for private funded certificates
```
ChrisJWoodcock, line 67: What's going on here?

src/SFA.DAS.AssessorService.Application/Handlers/Staff/StartCertificateHandler.cs
```diff
@@ -55,6 +55,8 @@ public async Task<Certificate> Handle(StartCertificateRequest request, Cancellat
                     certData.LearningStartDate = ilr.LearnStartDate;
                     certData.FullName = $"{ilr.GivenNames} {ilr.FamilyName}";
                     certificate.CertificateData = JsonConvert.SerializeObject(certData);
+                    
+                    // To be updated based on ILR Funding Model
```
ChrisJWoodcock, line 59: When will it be updated?

src/SFA.DAS.AssessorService.Web/ViewModels/Certificate/CertificateBaseViewModel.cs
```diff
@@ -20,6 +20,7 @@ public virtual void FromCertificate(Domain.Entities.Certificate cert)
             Level = CertificateData.StandardLevel;
             Uln = cert.Uln.ToString();
             IsPrivatelyFunded = cert.IsPrivatelyFunded;
```
ChrisJWoodcock, line 22: Why is this still here?

src/SFA.DAS.AssessorService.Application/Handlers/Search/SearchResultExtensions.cs
CodeWolf86:
```if (request.IsPrivatelyFunded)
                {
                    var toBeApprovedLogEntry = certificateLogs.FirstOrDefault(l => l.Status == CertificateStatus.ToBeApproved);
                    if (toBeApprovedLogEntry == null)
                        continue;
                    submittedLogEntry = toBeApprovedLogEntry;
                }```
                
This was the original logic. Privately Funded is no longer part of a search result, so wasn't sure how to re-work the logic. If it was privately funded and the to be approved existed it went over submitted, is it best just to take it out do you think?

File: src/SFA.DAS.AssessorService.Application/Handlers/Staff/StartCertificateHandler.cs (hunk 3 above)
Comment (CodeWolf86) [reply src/SFA.DAS.AssessorService.Application/Handlers/Staff/StartCertificateHandler.cs:59]: I was attempting to keep this PR as just the removal of privately funded without adding new stuff in to try and keep the commits clean. The SV-456 branch it's going into is our Epic branch for developing SV Record an Outcome so it will be updated there with another PR before merging into master.

File: src/SFA.DAS.AssessorService.Web/ViewModels/Certificate/CertificateBaseViewModel.cs (hunk 4 above)
Comment (CodeWolf86) [reply src/SFA.DAS.AssessorService.Web/ViewModels/Certificate/CertificateBaseViewModel.cs:22]: This plus your comments on the GetContactPreviousAddress are all interlinked. I originally dropped it from the base view model, but then there was no way to supply it into the GetContactPreviousAddress method. Which behind the scenes queries into this:

```
            var certificateAddress = await (from certificateLog in _context.CertificateLogs
                join certificate in _context.Certificates on certificateLog.CertificateId equals certificate.Id
                where statuses.Contains(certificate.Status) && certificateLog.Username == userName
                                                            && certificate.IsPrivatelyFunded == isPrivatelyFunded
                let certificateData = JsonConvert.DeserializeObject<CertificateData>(certificate.CertificateData)
```

I want to delete the last restriction clause when searching for a certificate address, which would remove the items you've highlighted. I'm unsure of the impact this has.

src/SFA.DAS.AssessorService.Application/Handlers/Staff/StartCertificateHandler.cs
CodeWolf86: I've updated it here so it's not lost, it's a simple change.

src/SFA.DAS.AssessorService.Web/ViewModels/Certificate/CertificateBaseViewModel.cs
CodeWolf86: I've removed privately funded, looking further i don't think it makes any sense to keep it in here it doesn't add value.

src/SFA.DAS.AssessorService.Application.Api.Client/Clients/CertificateApiClient.cs
CodeWolf86: Addressed in latest commit.

PR #1134, opened 2021-03-30
SV-568 handle version number as string in assessor service

ben1stone:
Temporarily based on EpaoStandardsImport to allow updates to Get Standard Version Options and import to be tested but should not be merged as part of that release


CodeWolf86:
Has been superceded by : https://github.com/SkillsFundingAgency/das-assessor-service/pull/1309

PR #1136, opened 2021-03-31
CON-2910 Sanitise certificates whilst adding; return all certificates for batch

ChrisJWoodcock:
Sanitise certificates whilst adding; return all certificates for batch numbers correctly.


src/SFA.DAS.AssessorService.Data/CertificateBatchLogRepository.cs
```diff
@@ -28,6 +31,65 @@ public async Task<CertificateBatchLog> GetCertificateBatchLog(string certificate
             return certificateBatchLog;
         }
 
+        public async Task<List<CertificatePrintSummary>> GetCertificatesForBatch(int batchNumber)
```
cofaulco, line 34: how come we have moved this method to the `CertificateBatchRepository.cs` I think it made more sense on the `CertificateRepository.cs` ?

cofaulco (commented):
1 comment

src/SFA.DAS.AssessorService.Data/CertificateBatchLogRepository.cs
ChrisJWoodcock: It operates mainly on the [CertificateBatchLogs] table as it now correctly gets those certificates which were actually included in a given batch rather than those which are currently in a given batch.

src/SFA.DAS.AssessorService.Data/CertificateRepository.cs
```diff
@@ -291,14 +294,16 @@ public async Task<int> GetCertificatesReadyToPrintCount(string[] excludedOverall
...
             var certificateIds = await _unitOfWork.Connection.QueryAsync<Guid>(
-                "SELECT TOP(@numberOfCertifictes)" +
-                    "c.[Id] " +
-                "FROM" + 
-                    "[Certificates] c " + 
-                "WHERE " +
-                    "JSON_VALUE(CertificateData, '$.OverallGrade') NOT IN @excludedOverallGrades " +
-                    "AND c.Status IN @includedStatus " +
-                    "AND BatchNumber IS NULL",
+              @"SELECT TOP(@numberOfCertifictes)
```
cofaulco, line 297: typo here

cofaulco (commented):
1 typo

cofaulco (approved):

src/SFA.DAS.AssessorService.Data/CertificateRepository.cs
ChrisJWoodcock: Done

PR #1139, opened 2021-04-07
CON-3361 Display the username of the submit action instead of submitted status

ChrisJWoodcock:


src/SFA.DAS.AssessorService.Application/Handlers/Certificates/GetCertificatesHistoryHandler.cs
```diff
@@ -71,7 +71,7 @@ private async Task<PaginatedList<CertificateSummaryResponse>> MapCertificates(Pa
                         recordedBy = certificate.CertificateLogs
                             .OrderByDescending(q => q.EventTime)
                             .FirstOrDefault(certificateLog =>
-                                certificateLog.Status == Domain.Consts.CertificateStatus.Submitted)?.Username;
+                                certificateLog.Action == Domain.Consts.CertificateActions.Submit)?.Username;
```
cofaulco, line 74: no existing unit tests around this area?

cofaulco (commented):
1 comment

src/SFA.DAS.AssessorService.Application/Handlers/Certificates/GetCertificatesHistoryHandler.cs
ChrisJWoodcock: I have added a unit test for this scenario

cofaulco (approved):

PR #1142, opened 2021-04-08
SV-576 Create Standard Version Selection Page

CodeWolf86:


src/SFA.DAS.AssessorService.Web/Controllers/CertificateVersionController.cs
```diff
@@ -122,9 +122,16 @@ private async Task<IActionResult> SaveViewModel(CertificateVersionViewModel vm,
             }
             var certSession = JsonConvert.DeserializeObject<CertificateSession>(sessionString);
 
+            if(string.IsNullOrWhiteSpace(vm.StandardUId))
+            {
+                ModelState.AddModelError("", "A version must be selected");
+            }
+
```
ben1stone, line 129: Can this be added to a separate view model validator class [like these](https://github.com/SkillsFundingAgency/das-assessor-service/tree/master/src/SFA.DAS.AssessorService.Web/Validators)
I think that may help the validation summary box link as the message will be attached to a property

ben1stone (approved):

src/SFA.DAS.AssessorService.Web/Controllers/CertificateVersionController.cs
CodeWolf86: Yep have added in the validator as it uses that as a fallback if JS fails

PR #1145, opened 2021-04-09
SV-577 Update options page content

ben1stone:


src/SFA.DAS.AssessorService.Web/Views/Certificate/Option.cshtml
```diff
@@ -47,10 +47,8 @@ else
                 }
 
                 <h1 class="govuk-heading-xl">
-                    @Localizer["Heading"]
+                    Choose the option for @Model.Standard
```
CodeWolf86, line 50: You can keep the Localizer in like here: https://github.com/SkillsFundingAgency/das-assessor-service/blob/SV-456-StandardVersionRecordOutcome/src/SFA.DAS.AssessorService.Web/Views/Certificate/Version.cshtml#L50

CodeWolf86 (changes requested):

CodeWolf86 (approved):

PR #1148, opened 2021-04-12
SV-584-Implement Standard Version Validation For EPAO's not allowed to assess that version

CodeWolf86:


src/SFA.DAS.AssessorService.Web.UnitTests/CertificateTests/CertificateVersionControllerTests.cs
```diff
@@ -34,6 +34,7 @@ public class CertificateVersionControllerTests
         private const int Ukprn = 123456;
         private const string Username = "TestProviderUsername";
         private Guid CertificateId = Guid.NewGuid();
+        private const string EpaoId = "EpaoId1234";
```
ChrisJWoodcock, line 37: Doesn't really matter but this would typically be like EPA0123 as that is the actual format

src/SFA.DAS.AssessorService.Web/Controllers/CertificateVersionController.cs
```diff
@@ -132,8 +133,19 @@ private async Task<IActionResult> SaveViewModel(CertificateVersionViewModel vm,
             var standardVersionChanged = vm.StandardUId != certificate.StandardUId;
             var standardVersion = await _standardVersionClient.GetStandardVersionByStandardUId(vm.StandardUId);
 
+            // Retrieve what versions the EPAO is able to assess
```
ChrisJWoodcock, line 136: Unlike most other controllers the set of Certificate controllers does use FluentValidation - so this should use it too - to maintain that 'consistency'

ChrisJWoodcock (changes requested):

src/SFA.DAS.AssessorService.Application.Api.Client/Clients/StandardVersionClient.cs
```diff
@@ -39,12 +39,31 @@ public async Task<StandardVersion> GetStandardVersionByStandardUId(string standa
                 return await RequestAndDeserialiseAsync<StandardVersion>(request, $"Could not find the standard {standardUId}");
             }
         }
+
+        public async Task<IEnumerable<StandardVersion>> GetEpaoRegisteredStandardVersions(string epaoOrgId)
```
alandburns, line 43: If I'm being pedantic "epaoOrgId" should be "epaOrgId" as in End Point Assessor Organisation Id.

CodeWolf86: Have renamed the variable where it was incorrect 👍

src/SFA.DAS.AssessorService.Web.UnitTests/CertificateTests/CertificateVersionControllerTests.cs
CodeWolf86: Have updated the value to be a better reflection 👍

src/SFA.DAS.AssessorService.Web/Controllers/CertificateVersionController.cs
ChrisJWoodcock: There are issues with the existing code to moving this validation into a 'FluentValidation' rule, the session state and the claim being two of them - therefore the fluent validation will not be used in this case and a tech debt will be raised instead.

ChrisJWoodcock (approved):

src/SFA.DAS.AssessorService.Web/Controllers/CertificateVersionController.cs
CodeWolf86: Tech debt ticket added to Standard Versions tech debt epic: https://skillsfundingagency.atlassian.net/browse/SV-603

PR #1149, opened 2021-04-13
SV-604 Update Assessment Check Page

CodeWolf86:


src/SFA.DAS.AssessorService.Web.UnitTests/CertificateTests/CertificateControllerTests.cs
```diff
@@ -83,16 +93,24 @@ public async Task WhenStartingANewCertificate_WithSingleVersionNoOptions_GoesToD
...
+                 });
             var result = await _certificateController.Start(model) as RedirectToActionResult;
 
             _mockCertificateApiClient.Verify(s => s.Start(It.Is<StartCertificateRequest>(
                 m => m.StandardCode == model.StdCode && m.Uln == model.Uln
                 && m.Username == Username && m.UkPrn == Ukprn)));
 
-            _mockSessionService.Verify(c => c.Set("CertificateSession", It.Is<CertificateSession>(
-                s => s.CertificateId == CertificateId && s.Uln == model.Uln && s.StandardCode == model.StdCode)));
+            _mockSessionService.Verify(c => c.Set("CertificateSession", It.Is<CertificateSession>(s => s.CertificateId == CertificateId && s.Uln == model.Uln && s.StandardCode == model.StdCode)));
```
cofaulco, line 112: `setSession` which is set in the above call back can have its values asserted rather than this verify. Much cleaner, and more straightforward when something goes wrong

CodeWolf86: Yep that's fair. I was fighting with the verify to try and get it to verify the versions and so introduced the callback to capture correctly to allow an assert. Just didn't take it all the way and just remove the verify. Simplifies the multiple checks on the set to just confirming the final session object which simplifies the test.

src/SFA.DAS.AssessorService.Web.UnitTests/CertificateTests/CertificateVersionControllerTests.cs
```diff
@@ -117,13 +136,13 @@ public async Task WhenSelectingAVersion_WhenLoadingModel_WithOneVersion_Redirect
         }
 
         [Test, MoqAutoData]
-        public async Task WhenSelectingAVersion_WhenLoadingModel_WithOneVersion_RedirectsToOptionPage(CertificateSession session, StandardVersion standardVersion, StandardOptions options)
+        public async Task WhenSelectingAVersion_WhenLoadingModel_WithOneVersion_RedirectsToOptionPage(CertificateSession session, SharedModelTypes.StandardVersion standardVersion, StandardOptions options)
```
cofaulco, line 139: Not immediately obvious from the name of this test that it is different from the one above. Might be worth including `_WithOptions`

alandburns (approved):
Looks good to me.
Now handles the partial completed Certs, with changing Options - with/without changing version.

cofaulco (approved):

src/SFA.DAS.AssessorService.Web.UnitTests/CertificateTests/CertificateControllerTests.cs
```diff
@@ -64,17 +64,29 @@ public void SetUp()
         [Test, MoqAutoData]
         public async Task WhenStartingANewCertificate_WithSingleVersionNoOptions_GoesToDeclaration(CertificateStartViewModel model, StandardVersion standard)
         {
+            CertificateSession setSession = new CertificateSession();
             _mockStandardVersionClient.Setup(s => s.GetStandardVersionsByLarsCode(model.StdCode)).ReturnsAsync(new List<StandardVersion> { standard });
             _mockStandardServiceClient.Setup(s => s.GetStandardOptions(standard.StandardUId)).ReturnsAsync(new StandardOptions());
+            _mockSessionService.Setup(c => c.Set("CertificateSession", It.IsAny<object>()))
+                .Callback<string, object>((key, session) =>
+                {
+                    if (key == "CertificateSession" && session is CertificateSession)
```
ChrisJWoodcock, line 73: Good opportunity to clean the code a bit could use nameof(CertificateSession)

src/SFA.DAS.AssessorService.Web/Controllers/SearchController.cs
```diff
@@ -98,7 +99,7 @@ private void GetSelectedStandardViewModel(SearchRequestViewModel result)
                 ShowExtraInfo = resultViewModel.ShowExtraInfo,
                 UlnAlreadyExists = resultViewModel.UlnAlreadyExists,
                 IsNoMatchingFamilyName = resultViewModel.IsNoMatchingFamilyName,
-                Versions = resultViewModel?.Versions.Select(v => (StandardVersion)v)
+                Versions = resultViewModel?.Versions
```
ChrisJWoodcock, line 102: can resultViewModel be null - if it can this check is good but why no check above?

src/SFA.DAS.AssessorService.Web/ViewModels/Shared/StandardVersion.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace SFA.DAS.AssessorService.Web.ViewModels.Shared
```
ChrisJWoodcock, line 1: All the other class names in the ViewModels are ClassNameViewModel

src/SFA.DAS.AssessorService.Web/ViewModels/Shared/StandardVersion.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    public class StandardVersion
+    {
+        public string Title { get; set; }
+        public string StandardUId { get; set; }
+        public string Version { get; set; }
+
+        public static implicit operator StandardVersion(Api.Types.Models.Standards.StandardVersion version)
```
ChrisJWoodcock, line 9: We usually use mappers for this sort of thing - I don't think this type of conversion is used anyway else in the code

PR #1150, opened 2021-04-14
SV-586 implement back button logic

CodeWolf86:
This PR is for the implementation of the back logic in the record a grade journey. The pages that occur in order when performing a record a grade are:

- Search Result -> Select Version* -> Select Option* -> Declaration -------------> Check Page

The two pages marked with an asterisk are optional. As a result:
- When the user is on the declaration page, they can return to Version, Option or Result.
- When the user is on the Option Page, they can return to Version or Result
- When the user in on the Version page, they can only return to search result.

In addition, there is the Check Page to consider, where on final checking of the grade, the user can return to Version or the Option page independently.
- If the user returns to the option page, when they click back it should return to the check page
- If the user returns to the version page, when they click back it should return to the check page
- if the user returns to the version page, and doesn't change the version and clicks continue when the standard has options, it should return to the check page
- if the user returns to the version page, and does change the version and clicks continue it should go to the options page as options can change between versions. Upon continuing it should return to the check page.
- if the user returns to the version page, changes the version, continues to the option page, then clicks back, it should return to the version page, not the check page, and subsequently if they click back again, back to the check page.

These changes are encompassed in the first commit of the PR for ease of reviewing. The subsequent commits are to address comments on a previous PR which were missed due to an earlier merge [here](https://github.com/SkillsFundingAgency/das-assessor-service/pull/1149).

These changes were performed after this PR commit as there was crossover and it was deemed easier to resolve in one go. These changes include the below. Please bear that in mind when reviewing the original commit before these improvements.
- Changing the string "CertificateSession" to nameof "CertificateSession"
- Removing an erroneous null check after changing the versions field population in the search handler
- Renaming the StandardVersion model to StandardVersionViewModel - Thus simplifying naming of POCOS across the files
- Removing an Explicit Mapper from StandardVersion Api type to an AutoMapper Profile


src/SFA.DAS.AssessorService.Web/Controllers/CertificateOptionController.cs
```diff
@@ -135,5 +134,62 @@ private string GetModelValues<T>(T viewModel)
...
+                //redirectToVersion is where a user has gone to the check page, selected change version
+                //has changed version, then has to pick a new option as well if they exist
+                //as a result redirecttocheck is passed so on selection of option the user goes back to the check page
+                //however if a user selects back, they must go back to the version select page.
+                //If a user on the check page changes just the option, then that should just go 
+                //back to the check page.
+                //redirectfromversion is always set when coming from version page to allow for this
+                //however it's removed beyond that page, therefore if coming back from declaration
+                //the final version count check allows that fallback.
+                var redirectToCheck = SessionService.Exists("redirecttocheck") && bool.Parse(SessionService.Get("redirecttocheck"));
```
cofaulco, line 164: this seems like it could be encapsulated in a `SessionService.TryGet<bool>("redirecttocheck")` ?

CodeWolf86: I've implemented a version of TryGet to simplify this code.

src/SFA.DAS.AssessorService.Web/Controllers/CertificateOptionController.cs
```diff
@@ -25,24 +25,24 @@ public CertificateOptionController(ILogger<CertificateController> logger, IHttpC
...
-            
-            return await LoadViewModel("~/Views/Certificate/Option.cshtml", isFromStandard);
+
+            return await LoadViewModel("~/Views/Certificate/Option.cshtml");
         }
 
-        private async Task<IActionResult> LoadViewModel(string view, bool? isFromStandard)
+        private async Task<IActionResult> LoadViewModel(string view)
         {
             var username = ContextAccessor.HttpContext.User.FindFirst("http://schemas.xmlsoap.org/ws/2005/05/identity/claims/upn")?.Value;
```
ChrisJWoodcock, line 41: This could be moved to the base class it is all over the place the same

src/SFA.DAS.AssessorService.Web/Controllers/CertificateOptionController.cs
```diff
@@ -54,7 +54,7 @@ private async Task<IActionResult> LoadViewModel(string view, bool? isFromStandar
             else
                 SessionService.Remove("redirecttocheck");
                 
-            var sessionString = SessionService.Get("CertificateSession");
+            var sessionString = SessionService.Get(nameof(CertificateSession));
```
ChrisJWoodcock, line 57: There is very similar code to this in lots of places perhaps a generic method on the base class, the different seems to be the name of the view model

ChrisJWoodcock (approved):
The comments are nice but not must do's

src/SFA.DAS.AssessorService.Web/Controllers/CertificateOptionController.cs
ChrisJWoodcock: That was supposed to be line 41

src/SFA.DAS.AssessorService.Web/Controllers/CertificateOptionController.cs
ChrisJWoodcock: And this was supposed to be about checking for the empty session - I don't know why the comments are not on the correct lines

src/SFA.DAS.AssessorService.Web/Controllers/CertificateOptionController.cs
CodeWolf86: Have refactored into base class 👍

src/SFA.DAS.AssessorService.Web/Controllers/CertificateOptionController.cs
CodeWolf86: Have also refactored into base class 👍

PR #1158, opened 2021-04-20
SV-609 confirm and submit options validation

ben1stone:
I've added a validation check for the Option property on the submit page to stop errors when changing the version but then going back to the check page before selecting a new option.

I've mapped the existing TempData["HideOption"] to a view model property so that I can conditionally validate Option when the current selected Standard version has options.


src/SFA.DAS.AssessorService.Web/Views/Certificate/Check.cshtml
```diff
@@ -41,6 +42,7 @@ else
                     <div class="govuk-error-summary__body">
                         <ul class="govuk-list govuk-error-summary__list">
                             <li><a sfa-validation-for="Name"></a></li>
+                            <li><a sfa-validation-for="Option"></a></li>
```
alandburns, line 45: Option should be higher up the list - the otehr fields all relate to the postal address details

src/SFA.DAS.AssessorService.Web/Views/Certificate/Check.cshtml
```diff
@@ -6,6 +6,7 @@
 
 @{
     ViewBag.Title = (!ViewData.ModelState.IsValid ? "Error: " : "") + "Check and approve the assessment details";
+    Model.StandardHasOptions = (bool)TempData["HideOption"] == false;
```
CodeWolf86, line 9: I would rework this. Rather than copy the temp data into the model, just use the model.

`            var sessionString = SessionService.Get(nameof(CertificateSession));
            if (sessionString == null)
            {
                return RedirectToAction("Index", "Search");
            }
            var certSession = JsonConvert.DeserializeObject<CertificateSession>(sessionString);
            TempData["HideOption"] = certSession.Options == null || !certSession.Options.Any();
            TempData["HideChangeOption"] = certSession.Options == null || certSession.Options.Count <= 1;
            TempData["HideChangeVersion"] = certSession.Versions == null || certSession.Versions.Count <= 1;`

Migrate the above into the LoadViewModel method, as that already also retrieves the session. Create Model properties for all 3 of the above ( so 2 more than what you have ) then just set that on the model rather than temp data, and switch the view to use the model properties. 

Feels like it would simplify the overall approach?

src/SFA.DAS.AssessorService.Web.UnitTests/CertificateTests/When_submitting_passed_certificate_with_no_option_selected.cs
```diff
@@ -0,0 +1,71 @@
...
+using Organisation = SFA.DAS.AssessorService.Domain.Entities.Organisation;
+
+namespace SFA.DAS.AssessorService.Web.UnitTests.CertificateTests
+{
+    public class When_submitting_passed_certificate_with_no_option_selected : CertificateCheckControllerTestBase
+    {
+        private ViewResult _result;
+
+        [SetUp]
+        public void Arrange()
```
CodeWolf86, line 24: I really don't like all the setup and config in the existing unit test arrangements. When I've been making changes, i've been standardising the unit tests, see the CertificateOptionControllerTests.cs or CertificateVersionControllerTests.cs

Can we restructure the check controller tests into that format? Do you prefer this format? or the one i've suggested above, happy to talk it through

CodeWolf86 (changes requested):

src/SFA.DAS.AssessorService.Web/Controllers/CertificateBaseController.cs
```diff
@@ -51,6 +51,7 @@ protected async Task<IActionResult> LoadViewModel<T>(string view) where T : Cert
             Logger.LogInformation($"Got Certificate for {typeof(T).Name} requested by {username} with Id {certificate.Id}");
 
             viewModel.FromCertificate(certificate);
+            viewModel.SetStandardHasVersionsAndOptions(certSession);
```
CodeWolf86, line 54: This is getting there but I don't think the right idea is to add properties to the base view model and the base controller when it's only used in the check controller.

How about making the LoadViewModel method virtual in the base and overriding in the check controller to call base + do the additional work?

CodeWolf86 (changes requested):

CodeWolf86 (approved):

PR #1159, opened 2021-04-21
SV-595 Add in Options Cache And Transpose Options to Standard Version Controller

CodeWolf86:
This has been reworked in the PR such that upon the import running we now store the options in the new StandardOption table.

This has then been transposed upwards and removed the need for calling GetOptions on the OuterApi or variations therein.

As part of this PR, the Options endpoints have been moved into the StandardVersion controller to encapsulate all of the standards version work and to allow the eventual removal of the Standard-Service Controller which will be replaced by the Standard-Version Controller.

As Standard Versions proceeds through the application, the StandardCollation and Options Tables should be able to be removed, but this will only occur once all the api calls under the hood use the new standards and options cache as opposed to the older one.

Tech Debt Ticket also raised to remove option endpoints at a future point as now redundant.
[Tech Debt Ticket](https://skillsfundingagency.atlassian.net/browse/SV-634)


alandburns (commented):
Looks good. I will run a test.

alandburns (approved):

ben1stone (approved):

PR #1161, opened 2021-04-22
SV-511 allow resuming when recording a retake after fail

ben1stone:
We were previously unable to leave and resume recording a pass from a previously failed apprenticeship as when Starting the new certificate if a failed certificate was found it would be returned and we were allowed to overwrite it. However, the status of the certificate was still Submitted, so when we changed the grade to a pass and left the search found a passing, submitted certificate and did not allow changes to be made.

I have added a check when starting a new certificate so that if the result is Submitted + Fail then set status to draft and log that it is recording a retake. Submitted is important so draft fails aren't logged as retakes.


CodeWolf86 (approved):

alandburns (approved):
Test locally.

PR #1165, opened 2021-04-26
SV-619 validate all fields on check page on submit

ben1stone:
I've added validation for Grade, Version and Achievement date. 

The existing unit tests for the CertificateCheckController were getting complicated because it was loading different certificates for each scenario and used a concrete validator. I've now changed the controller tests to use a mocked validator to check if the redirection and update API calls are correct if the view model is valid or not and then added a set of tests on the validator itself.


alandburns (approved):
Tested locally.

PR #1166, opened 2021-04-27
SV-476 record a grade fail to pass 1 standard many versions with options

ben1stone:
I've included the certificate logs in with the GetCertificate by certificateId call and added a property to the CertificateVersion ViewModel.


ben1stone:
Closed, these changes are now included in the main story pr SV-455

PR #1167, opened 2021-04-28
SV-455 record pass from previous fail

ben1stone:
Some of this work (SV-477) has already been reviewed by Alan but I have merged a few subtask tickets into this story branch. Some scenarios did not need any changes.


src/SFA.DAS.AssessorService.Domain/Extensions/DateTimeUtils.cs
```diff
@@ -63,5 +63,10 @@ public static string ToSfaShortestDateString(this DateTime? time)
             return time == null ? 
                 string.Empty : time.Value.ToString("dd MMM yy");
         }
+
+        public static string ToNumericalDateString(this DateTime? time)
+        {
+            return time == null ? string.Empty : time.Value.ToString("dd/MM/yyyy");
```
CodeWolf86, line 69: You will probably be able to get away with `time.Value.ToString("d")`  https://docs.microsoft.com/en-us/dotnet/api/system.datetime.tostring?view=net-5.0

src/SFA.DAS.AssessorService.Web/Views/Search/Result.cshtml
```diff
@@ -28,7 +28,7 @@
                                 <dd class="govuk-summary-list__value">@Model.LearnerStartDate</dd>
                             </div>
                         </dl>
-                        <p class="govuk-body">If the name, standard or level is incorrect, you need to agree on the correct details along with the training provider and the employer. The training provider needs to then update the Individualised Learner Record (<abbr title="Individualised Learner Record">ILR</abbr>).</p>
+                        <p class="govuk-body">If the name, standard or level is incorrect, you need to agree on the correct details with the training provider and employer. Once the training provider needs to then updates the Individualised Learner Record (<abbr title="Individualised Learner Record">ILR</abbr>), return here and continue.</p>
```
CodeWolf86, line 31: `Once the training provider needs to then updates` ? may need rewording

src/SFA.DAS.AssessorService.Web/Views/Search/Result.cshtml
```diff
@@ -53,7 +53,7 @@
                                 @(await Html.PartialAsync($"./_ApprenticeshipExtraInfoListRows.cshtml", Model))
                             }
                         </dl>
-                        <p class="govuk-body">If the name, standard or level is incorrect, you need to agree on the correct details along with the training provider and the employer. The training provider needs to then update the Individualised Learner Record (<abbr title="Individualised Learner Record">ILR</abbr>).</p>
+                        <p class="govuk-body">If the name, standard or level is incorrect, you need to agree on the correct details with the training provider the employer. Once the training provider updates the Individualised Learner Record (<abbr title="Individualised Learner Record">ILR</abbr>), return here and continue.</p>
```
CodeWolf86, line 56: `with the training provider the employer` change `the` to `and` as above?

CodeWolf86 (changes requested):

CodeWolf86 (approved):

src/SFA.DAS.AssessorService.Domain/Extensions/DateTimeUtils.cs
alandburns: I prefer the explicit "dd/MM/yyyy" to avoid any unintentional "en-US" defaults. But don't we have a SFA date format.
Yes. we have src\SFA.DAS.AssessorService.Domain\Extensions\DateTimeUtils.cs

CodeWolf86: That's fair. @ben1stone may be worth either changing back or enforcing the en culture in the overload. Will leave it up to you, I know the rest of the dates are explicit in this util class.

alandburns: I am also concerned that we are being asked to format a date in the Assessor service UI in a different way.
I rather suspect that this is a CONTENT mis-align, as I don't see why we should be adding a new format type to display date.

ben1stone: @alandburns In the main summary lists the date is 1 January 2021 but this was inside a dropdown so thought that was the distinction but I'll check with UX

PR #1169, opened 2021-04-30
SV-611 search result page approved for standard version

ben1stone:


src/SFA.DAS.AssessorService.Application/Handlers/Search/SearchHandler.cs
```diff
@@ -111,10 +110,12 @@ private async Task<List<SearchResult>> Search(SearchQuery request, CancellationT
 
         private async Task<List<int>> GetEpaoStandards(Organisation thisEpao)
```
CodeWolf86, line 111: Can we rename the method? `GetApprovedStandardsForEpao`? `GetEpaoApprovedStandardsWithAtLeastOneVersion` ? I'm trying to make it more descriptive to say what it's doing, GetEpaoStandards does say what it's doing but not the whole story.

CodeWolf86 (approved):

ben1stone:
@alandburns I've updated to account for the scenario you described on the Jira ticket comment. I've added if's in the view around the certificate data fields because these are filled during the mapping if it is the correct EPAO so if it is a different EPAO they will be null

PR #1172, opened 2021-05-04
SV-501 Create EPA Post Record

CodeWolf86:


src/SFA.DAS.AssessorService.Application.UnitTests/Handlers/Certificates/StartCertificateHandlerTests/When_called_and_no_existing_certificate.cs
```diff
@@ -25,16 +26,17 @@ public class When_called_and_no_existing_certificate
...
         [SetUp]
         public void Arrange()
         {
             _certificateRepository = new Mock<ICertificateRepository>();
             _certificateRepository.Setup(r => r.GetCertificate(1111111111, 30)).ReturnsAsync(default(Certificate));
 
+            _savedCertificates = new List<Certificate>();
             _certificateRepository.Setup(r => r.New(It.IsAny<Certificate>())).ReturnsAsync(new Certificate() { CertificateReferenceId = 10000 })
-                .Callback<Certificate>(c => _savedCertificate = c);
+                .Callback<Certificate>(c => _savedCertificates.Add(c));
```
ben1stone, line 40: Just wondering what this setup .Callback does?

src/SFA.DAS.AssessorService.Application.UnitTests/Handlers/Certificates/StartCertificateHandlerTests/When_called_and_no_existing_certificate.cs
```diff
@@ -95,6 +115,18 @@ public void Arrange()
...
+                        StandardCode = 30,
+                        UkPrn = 88888888,
+                        Uln = 333333333,
+                        Username = "user",
+                        CourseOption = "CourseOption",
+                        StandardUId = "ST0016",
+                    }, new CancellationToken()).Wait();
+
             _returnedCertificate = _startCertificateHandler
                 .Handle(
```
ben1stone, line 131: Should these Handler calls be in their own specific tests rather than 3 in the setup? Also, the last call assigns the certificate to the `_returnedCertificate` field but the field isn't being checked against it just Verifies the repository method.

src/SFA.DAS.AssessorService.Data.IntegrationTests/StandardsGetTests.cs
```diff
@@ -43,7 +43,7 @@ public async Task GetStandardByStandardReferenceAndVersion_ReturnsCorrectStandar
         {
             var expectedStandard = _standards.Single(s => s.StandardUId == standardUId);
 
-            var standard = await _repository.GetStandardByStandardReferenceAndVersion(standardReference, version);
+            var standard = await _repository.GetStandardVersionByIFateReferenceNumber(standardReference, version);
 
             Assert.AreEqual(expectedStandard.StandardUId, standard.StandardUId);
         }
```
ben1stone, line 49: Does this test need a case where it gets the correct standard when no version is supplied?

src/SFA.DAS.AssessorService.Data/StandardRepository.cs
```diff
@@ -240,7 +263,7 @@ private async Task<Standard> GetLatestVersionOfAStandardByIFateReferenceNumber(s
             return result;
         }
 
-        private async Task<Standard> GetStandardsByStandardReferenceAndVersionInternal(string standardReference, string version)
+        private async Task<Standard> GetStandardByIFateReferenceNumberAndVersionInternal(string ifateReferenceNumber, string version)
         {
             var sql = @"SELECT StandardUId, IfateReferenceNumber, LarsCode, Title, Version, Level, Status, TypicalDuration, 
                             MaxFunding, IsActive, LastDateStarts, EffectiveTo, EffectiveFrom, VersionEarliestStartDate,
```
ben1stone, line 269: Column names are in square brackets on some queries and not others

src/SFA.DAS.AssessorService.Data/StandardRepository.cs
```diff
@@ -249,7 +272,22 @@ private async Task<Standard> GetStandardsByStandardReferenceAndVersionInternal(s
...
+                param: new { ifateReferenceNumber, version },
+                transaction: _unitOfWork.Transaction);
+
+            return results.FirstOrDefault();
+        }
+
+        private async Task<Standard> GetStandardByLarsCodeAndVersionInternal(int larsCode, string version)
+        {
+            var sql = @"SELECT StandardUId, IfateReferenceNumber, LarsCode, Title, Version, Level, Status, TypicalDuration, 
+                            MaxFunding, IsActive, LastDateStarts, EffectiveTo, EffectiveFrom, VersionEarliestStartDate,
```
ben1stone, line 284: Column names are in square brackets on some queries and not others

ben1stone (changes requested):

src/SFA.DAS.AssessorService.Application.UnitTests/Handlers/Certificates/StartCertificateHandlerTests/When_called_and_no_existing_certificate.cs
CodeWolf86: I was holding off on reworking the test class but it's probably the right time to do it. I don't like the way it's structured so will change it to how we've upgraded them elsewhere. Watch this space.

src/SFA.DAS.AssessorService.Data/StandardRepository.cs
CodeWolf86: Have updated to keep it consistent

src/SFA.DAS.AssessorService.Data/StandardRepository.cs
CodeWolf86: Have updated to keep it consistent

src/SFA.DAS.AssessorService.Data.IntegrationTests/StandardsGetTests.cs
CodeWolf86: Have added a test case for version is empty to get the latest version

src/SFA.DAS.AssessorService.Application.UnitTests/Handlers/Certificates/StartCertificateHandlerTests/When_called_and_no_existing_certificate.cs
CodeWolf86: I have removed the horrible set of tests and then added in a new file that's hopefully neater

src/SFA.DAS.AssessorService.Application.UnitTests/Handlers/Certificates/StartCertificateHandlerTests/When_called_and_no_existing_certificate.cs
CodeWolf86: The Callback is invoked before or after the return depending on when it's added to the chain. The ideas is that if you wanted to do additional action in that window you are able to. In this case I wanted to capture the object passed into the method. I find it easier to read to save this object and assert on it, than a long winded verify though i think i've changed this now with the test rework

ben1stone: Thanks Chris!

ben1stone (approved):

PR #1174, opened 2021-05-06
SV-671 Hide form if all assessments are completed

ben1stone:
The diff isn't very clear on this but I have moved the start of the form to below the introductory information and applied a new AllAssessmentsCompleted property so the whole form is hidden instead of the radio buttons section. The AllAssessmentsCompleted is also used to provide the alternate title.


src/SFA.DAS.AssessorService.Web/Resources/Views/Search/ChooseStandard.en.resx
```diff
@@ -115,7 +115,7 @@
     <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <resheader name="writer">
-    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c5619PreviouslyRecordedAssessment34e089</value>
```
CodeWolf86, line 118: This a Typo?

CodeWolf86 (changes requested):

src/SFA.DAS.AssessorService.Web/Resources/Views/Search/ChooseStandard.en.resx
ben1stone: yep, fixed

alandburns (approved):
Tested locally.
Raised a separate issue on content consistency.

CodeWolf86 (approved):

PR #1179, opened 2021-05-07
SV-503 get certificate external api

ben1stone:


CodeWolf86 (changes requested):
Nice addition of tests. We don't have many of those for the external api at the moment. It's something i've skipped over so far as I think we need tests on the external api, and the internal api's external controllers especially around some of the mapping of fields using the profile.

As for this PR, I think there is still some work to do. The External Controller contains an `IsDraftCertificateDeemedAsReady` I think this needs to be checked if version has been set, along with if the option that's been set is valid, and there is an option if it's required. 

In addition on the internal Api controller - CertificateBatchController. It is making use of the StandardCollation tables which we are looking to remove as part of this work. ( See the BatchEpaController changes i've made for what I mean )

Also, the GetBatchCertificateRequestValidator also uses the same GetCollatedStandard methods so needs to be looked at.

One thing i've not checked is the GetEpaoRegisteredVersions, that stored procedure uses standard collation but that is something you can look at in this PR, or as a separate ticket as it's something I ignored initially on the EPA side from the sheer amount of changes.

Happy to talk this through when you're back.

src/SFA.DAS.AssessorService.Application.Api/Validators/ExternalApi/Certificates/GetBatchCertificateValidator.cs
```diff
@@ -18,20 +18,7 @@ public GetBatchCertificateRequestValidator(IStringLocalizer<GetBatchCertificateR
...
-                    {
-                        var collatedStandard = await standardService.GetStandard(m.StandardReference);
-                        if (m.StandardCode != collatedStandard?.StandardId)
-                        {
-                            context.AddFailure("StandardReference and StandardCode must be for the same Standard");
-                        }
-                    }
-                });
-            });
+            RuleFor(m => m.StandardCode).GreaterThan(0).WithMessage("Provide a valid Standard");
```
CodeWolf86, line 21: I'm debating this validation line. They provide a lars code or ifate reference which we use to get the latest standard. Then we populate both the StandardCode and StandardReference on the request. So in theory either they will both be null or both be set. 

I guess this is a catch all so that if it's not set, a valid id wasn't supplied.

src/SFA.DAS.AssessorService.Application.Api/Controllers/ExternalApi/CertificateBatchController.cs
```diff
@@ -23,13 +23,13 @@ namespace SFA.DAS.AssessorService.Application.Api.Controllers.ExternalApi
     public class CertificateBatchController : Controller
     {
         private readonly IMediator _mediator;
-        private readonly GetBatchCertificateRequestValidator _getValidator;
-        private readonly CreateBatchCertificateRequestValidator _createValidator;
-        private readonly UpdateBatchCertificateRequestValidator _updateValidator;
-        private readonly SubmitBatchCertificateRequestValidator _submitValidator;
-        private readonly DeleteBatchCertificateRequestValidator _deleteValidator;
+        private readonly IValidator<GetBatchCertificateRequest> _getValidator;
```
CodeWolf86, line 26: Nice rework to IValidator for testing

src/SFA.DAS.AssessorService.Application.Api.External/Helpers/CertificateHelpers.cs
```diff
@@ -17,6 +19,10 @@ public static bool IsDraftCertificateDeemedAsReady(Certificate certificate)
             {
                 return false;
             }
+            else if (certificate.CertificateData.LearningDetails.Version is null)
```
CodeWolf86, line 22: Good changes to make sure version is set along with option if applicable. Just a shame we don't tell them what's wrong with a certificate for submission. But that's beyond our scope

CodeWolf86 (approved):

PR #1183, opened 2021-05-12
Sv 503 get certificate external api

ben1stone:


src/SFA.DAS.AssessorService.Application.Api/Validators/ExternalApi/Certificates/GetBatchCertificateValidator.cs
```diff
@@ -26,48 +26,41 @@ public GetBatchCertificateRequestValidator(IStringLocalizer<GetBatchCertificateR
...
+                        if (sumbittingEpao is null)
                         {
                             context.AddFailure(new ValidationFailure("UkPrn", "Specified UKPRN not found"));
                         }
                         else
                         {
                             var providedStandards = await standardService.GetEpaoRegisteredStandards(sumbittingEpao.EndPointAssessorOrganisationId);
 
-                            if (!providedStandards.Any(s => s.StandardCode == m.StandardCode))
+                            if (!providedStandards.Any(s => s.StandardCode == m.StandardCode) && createdByCertificate == null )
```
CodeWolf86, line 43: What is the reasoning behind createdByCertificate being null along with not approved for the standard?

src/SFA.DAS.AssessorService.Application.Api/Validators/ExternalApi/Certificates/GetBatchCertificateValidator.cs
```diff
@@ -26,48 +26,41 @@ public GetBatchCertificateRequestValidator(IStringLocalizer<GetBatchCertificateR
...
-                            context.AddFailure(new ValidationFailure("Uln", "Cannot find apprentice with the specified Uln, FamilyName & Standard"));
-                        }
-                        else if (sumbittingEpao is null)
+                        if (sumbittingEpao is null)
                         {
                             context.AddFailure(new ValidationFailure("UkPrn", "Specified UKPRN not found"));
                         }
                         else
                         {
                             var providedStandards = await standardService.GetEpaoRegisteredStandards(sumbittingEpao.EndPointAssessorOrganisationId);
```
CodeWolf86, line 41: If we are stopping them from getting a certificate because they are not approved for that standard, should we also be validating on version as well?

src/SFA.DAS.AssessorService.Application.Api/Validators/ExternalApi/Certificates/GetBatchCertificateValidator.cs
```diff
@@ -26,48 +26,41 @@ public GetBatchCertificateRequestValidator(IStringLocalizer<GetBatchCertificateR
                 {
                     RuleFor(m => m).CustomAsync(async (m, context, cancellation) =>
                     {
-                        // NOTE: Currently we're making the Certificate & ILR record both mandatory
+                        // NOTE: Currently we're making the Certificate & ILR record both mandatory - this is wrong fixing it!
                         var requestedIlr = await ilrRepository.Get(m.Uln, m.StandardCode);
                         var sumbittingEpao = await organisationQueryRepository.GetByUkPrn(m.UkPrn);
```
CodeWolf86, line 31: slight typo here `sumbitting`

src/SFA.DAS.AssessorService.Application.Api/Validators/ExternalApi/Certificates/GetBatchCertificateValidator.cs
CodeWolf86, line 41: To rejig the code slightly, in the `else` of `if(submittingEpao is null` I would move some of the calls downwards. No point querying for stuff if the submitting Epao is null.
```
var createdByCertificateTask = certificateRepository.GetCertificateByOrgIdLastname(m.Uln, submittingEpao.EndPointAssessorOrganisationId, m.FamilyName);
                            var providedStandardsTask = standardService.GetEpaoRegisteredStandards(submittingEpao.EndPointAssessorOrganisationId);
                            var requestedIlrTask = ilrRepository.Get(m.Uln, m.StandardCode);
                            var existingCertificateTask = certificateRepository.GetCertificate(m.Uln, m.StandardCode);

                            await Task.WhenAll(createdByCertificateTask, providedStandardsTask, requestedIlrTask, existingCertificateTask);

                            var providedStandards = providedStandardsTask.Result;
                            var requestedIlr = requestedIlrTask.Result;
                            var createdByCertificate = createdByCertificateTask.Result;
                            var existingCertificate = existingCertificateTask.Result;```

**Review by CodeWolf86 — CHANGES_REQUESTED** [review]

File: src/SFA.DAS.AssessorService.Application.Api/Validators/ExternalApi/Certificates/GetBatchCertificateValidator.cs (hunk 1 above)
Comment (alandburns) [reply src/SFA.DAS.AssessorService.Application.Api/Validators/ExternalApi/Certificates/GetBatchCertificateValidator.cs:44]: What this is attempting to do - but please check my logic - is to allow certificates that were created by the EPAO, where that EPAO is no longer on the register for the standard. If the EPAO created the certificate then they can retrieve that data any time in the future.

File: src/SFA.DAS.AssessorService.Application.Api/Validators/ExternalApi/Certificates/GetBatchCertificateValidator.cs (hunk 2 above)
Comment (alandburns) [reply src/SFA.DAS.AssessorService.Application.Api/Validators/ExternalApi/Certificates/GetBatchCertificateValidator.cs]: The UI does not prevent the search returning results when the EPAO is not approved for the version recorded in the certificate. The check is that the EPAO knows the ULN, FamilyName and Standard, the combination of which is deemed sufficient to verify their legitimate access to the data, and that they are approved on that Standard (unless the EPAO was approved in the past when the certificate was created!)
The message returned if at least one of these fields is not matched to a learner is
"Cannot find apprentice with the specified Uln, FamilyName & Standard"

File: src/SFA.DAS.AssessorService.Application.Api/Validators/ExternalApi/Certificates/GetBatchCertificateValidator.cs [hunk 4]
```diff
@@ -26,48 +26,42 @@ public GetBatchCertificateRequestValidator(IStringLocalizer<GetBatchCertificateR
...
+                        if (submittingEpao is null)
                         {
                             context.AddFailure(new ValidationFailure("UkPrn", "Specified UKPRN not found"));
                         }
                         else
                         {
-                            var providedStandards = await standardService.GetEpaoRegisteredStandards(sumbittingEpao.EndPointAssessorOrganisationId);
+                            var requestedIlr = await ilrRepository.Get(m.Uln, m.StandardCode);
+                            var existingCertificate = await certificateRepository.GetCertificate(m.Uln, m.StandardCode);
+                            var createdByCertificate = await certificateRepository.GetCertificateByOrgIdLastname(m.Uln, submittingEpao.EndPointAssessorOrganisationId, m.FamilyName);
```
CodeWolf86, line 40: rename `createdByCertificate` to something like `existingCertificateCreatedByCallingEPAO` ? to make it obvious that this fact bypasses some validation

CodeWolf86 (changes requested):

src/SFA.DAS.AssessorService.Application.Api/Validators/ExternalApi/Certificates/GetBatchCertificateValidator.cs
alandburns: yep. Makes sense. Will you be making the change @ben1stone ?

CodeWolf86 (approved):

PR #1185, opened 2021-05-14
SV-507 external api get options

ben1stone:
I've added a new GetStandardOptionsForLatestStandardVersions to the internal API with a specific query, repository + service methods to return only the latest standard versions which have options. 

The External API was changed to call the new internal API method as GET ALL versions is not required for the external API but GET ALL is still used in the internal API

### DIFF
--- a/src/SFA.DAS.AssessorService.Application.Api.External/appsettings.json
+++ b/src/SFA.DAS.AssessorService.Application.Api.External/appsettings.json
@@ -13,5 +13,5 @@
   },
   "EnvironmentName": "LOCAL",
   "InstanceName": "LOCAL",
-  "UseSandboxServices": false
+  "UseSandboxServices": true
 }
--- a/src/SFA.DAS.AssessorService.Application.Api.External/appsettings.json
+++ b/src/SFA.DAS.AssessorService.Application.Api.External/appsettings.json
@@ -13,5 +13,5 @@
   },
   "EnvironmentName": "LOCAL",
   "InstanceName": "LOCAL",
-  "UseSandboxServices": true
+  "UseSandboxServices": false
 }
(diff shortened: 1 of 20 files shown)


src/SFA.DAS.AssessorService.Application.Api.External/appsettings.json
```diff
@@ -13,5 +13,5 @@
   },
   "EnvironmentName": "LOCAL",
   "InstanceName": "LOCAL",
-  "UseSandboxServices": false
+  "UseSandboxServices": true
```
alandburns, line 16: Is setting this to **true** deliberate?

src/SFA.DAS.AssessorService.Application.Api/Services/StandardService.cs
```diff
@@ -160,6 +160,22 @@ public async Task<IEnumerable<StandardOptions>> GetAllStandardOptions()
...
+        {
+            try
+            {
+                var options = await _standardRepository.GetStandardOptionsForLatestStandardVersions();
+
+                return options;
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, "STANDARD OPTIONS: Failed to get latest version of each standard");
```
CodeWolf86, line 173: Failed to get options for latest version of each standard?

CodeWolf86 (approved):

src/SFA.DAS.AssessorService.Application.Api/Controllers/StandardVersionController.cs
```diff
@@ -106,7 +117,7 @@ public async Task<IActionResult> GetStandardOptionsForStandard(string id)
         public async Task<IActionResult> GetStandardOptionsForStandardReferenceAndVersion(string standardReference, string version)
```
CodeWolf86, line 117: Should this method name and log message now be changed?

CodeWolf86 (approved):

PR #1194, opened 2021-05-21
SV-763 EPAChanged flag

techtoniq:


src/SFA.DAS.AssessorService.Database/Tables/Standards.sql
```diff
@@ -18,5 +18,6 @@
     [VersionLatestEndDate] DATETIME NULL, 
     [VersionApprovedForDelivery] DATETIME NULL, 
     [ProposedTypicalDuration] INT NOT NULL, 
-    [ProposedMaxFunding] INT NOT NULL
+    [ProposedMaxFunding] INT NOT NULL,
+	[EPAChanged] BIT NOT NULL
```
CodeWolf86, line 22: Just need to set a default of 0 so that the publish works and gets overridden the first time the import runs

CodeWolf86 (changes requested):

CodeWolf86 (approved):

PR #1204, opened 2021-05-27
CON-3540-EPAO Does not change once cert deleted

VasanthaKasirajan3008:


src/SFA.DAS.AssessorService.Application/Handlers/Staff/StartCertificateHandler.cs
```diff
@@ -43,26 +43,28 @@ public class StartCertificateHandler : IRequestHandler<StartCertificateRequest,
...
                     certData.LearnerFamilyName = ilr.FamilyName;
                     certData.LearningStartDate = ilr.LearnStartDate;
                     certData.FullName = $"{ilr.GivenNames} {ilr.FamilyName}";
                     certificate.CertificateData = JsonConvert.SerializeObject(certData);
                     certificate.IsPrivatelyFunded = ilr?.FundingModel == PrivateFundingModelNumber;
-
+                    certificate.OrganisationId = organisation.Id;
                     await _certificateRepository.Update(certificate, request.Username, null);
                 }
                 else if (certificate.Status == CertificateStatus.Submitted && certData.OverallGrade == CertificateGrade.Fail)
```
ChrisJWoodcock, line 70: New SV Code: Why does the Ilr details not need updating for this 'restart' case

src/SFA.DAS.AssessorService.Application/Handlers/Staff/StartCertificateHandler.cs
```diff
@@ -89,12 +91,10 @@ public async Task<Certificate> Handle(StartCertificateRequest request, Cancellat
...
 
-        private async Task<Certificate> CreateNewCertificate(StartCertificateRequest request)
+        private async Task<Certificate> CreateNewCertificate(StartCertificateRequest request, Organisation organisation)
         {
             _logger.LogInformation("CreateNewCertificate Before Get Ilr from db");
             var ilr = await _ilrRepository.Get(request.Uln, request.StandardCode);
-            _logger.LogInformation("CreateNewCertificate Before Get Organisation from db");
-            var organisation = await _organisationQueryRepository.GetByUkPrn(request.UkPrn);
             _logger.LogInformation("CreateNewCertificate Before Get Provider from API");
             var provider = await GetProviderFromUkprn(ilr.UkPrn);
```
ChrisJWoodcock, line 99: New SV Code: In the case of multiple standard version why is only the title of the first recorded on line 155 but for a single version all the properties are recorded on lines 137-148?

When a deleted certificate is re-used and in the mean time there may be additional standard versions should these versions which are stored in the certificate data not get refreshed?

ChrisJWoodcock: This is because when there are multiple versions the title is known but the other details will be supplied from the UI

src/SFA.DAS.AssessorService.Application/Handlers/Staff/StartCertificateHandler.cs
ChrisJWoodcock: This is because when passing a previously failed certificate the other details should remain the same

cofaulco (approved):

PR #1205, opened 2021-05-27
SV-525 add standard version to learner details query result

techtoniq:


src/SFA.DAS.AssessorService.Api.Types/Models/Staff/LearnerDetailResult.cs
```diff
@@ -11,6 +11,7 @@ public class LearnerDetailResult
         public string FamilyName { get; set; }
         public int StandardCode { get; set; }
         public string Standard { get; set; }
+        public string StandardVersion { get; set; }
```
CodeWolf86, line 14: Can we name the field `Version` ? just for consistency with elsewhere

src/SFA.DAS.AssessorService.Application/Handlers/Staff/GetLearnerDetailHandler.cs
```diff
@@ -86,6 +86,7 @@ public class GetLearnerDetailHandler : IRequestHandler<GetLearnerDetailRequest,
                 LearnStartDate = certificateData.LearningStartDate.HasValue ? certificateData.LearningStartDate : learner?.LearnStartDate,
```
CodeWolf86, line 86: Can we rework the handler a little bit while we are here?
Take out the `var standard = await _standardRepository.GetStandardCollationByStandardId(request.StdCode);`

Replace it with `GetStandardVersionByLarsCode(code,version)` then it's another collation call taken away. 

It would be nice if this had some unit tests but looking at it, there aren't any for the handlers as it stands currently, which is a shame. Given we want to get on with the admin stuff can you add a tech debt ticket to SV-363 to add Unit tests to the GetLearnerDetailsHandler?

src/SFA.DAS.AssessorService.Application/Handlers/Staff/GetLearnerDetailHandler.cs
```diff
@@ -78,15 +77,18 @@ public class GetLearnerDetailHandler : IRequestHandler<GetLearnerDetailRequest,
                 epao = await _organisationRepository.Get(certificate.OrganisationId) ?? new Organisation();
             }
 
+            var standard = await _standardRepository.GetStandardVersionByLarsCode(request.StdCode, certificateData.Version);
```
CodeWolf86, line 80: May want to add a null conditional onto certificateDate?.Version, as if the certificate is null you won't get the cert data.

CodeWolf86 (approved):

PR #1223, opened 2021-06-09
SV-456 sonar improvements

CodeWolf86:


src/SFA.DAS.AssessorService.Application.Api/Validators/ExternalApi/Epas/BatchEpaRequestValidator.cs
```diff
@@ -57,7 +57,7 @@ public BatchEpaRequestValidator(IStringLocalizer<BatchEpaRequestValidator> local
                     {
                         context.AddFailure(new ValidationFailure("CourseOption", "No course option available for this Standard and version. Must be empty"));
                     }
-                    else if (standardOptions != null && standardOptions.HasOptions() && standardOptions.CourseOption.All(a => a.IndexOf(m.CourseOption, StringComparison.OrdinalIgnoreCase) == -1))
+                    else if (standardOptions.HasOptions() && standardOptions.CourseOption.All(a => a.IndexOf(m.CourseOption, StringComparison.OrdinalIgnoreCase) == -1))
```
CodeWolf86, line 60: the if above checks if it's null, so by implicit logic, it's not null here so can be removed

src/SFA.DAS.AssessorService.Data/SandboxDbRepository.cs
```diff
@@ -221,15 +221,19 @@ private void BulkCopyData(SqlTransaction transaction, List<string> tablesToCopy)
                 foreach (var table in tablesToCopy)
                 {
                     _logger.LogDebug($"\tSyncing table: {table}");
-                    var idField = "[Id]";
+
```
CodeWolf86, line 224: Can you double check my working @alandburns  ?

CodeWolf86:
Original Analysis on the branch
image

PR #1225, opened 2021-06-12
Return versions in summary as CSV

techtoniq:
Apply.ApplyData.Versions is an object (array of strings).
ApplicationSummaryItem in the AssessorService API requires the "Versions" field to be a CSV string of version numbers if present.
This tweak to the stored function parses the array of strings from the json Versions field to return a CSV.
Not sure if there is a better way to unformat the json rather than the multiple REPLACEs.


techtoniq:
Ignore this, a new PR will be opened.

PR #1226, opened 2021-06-14
SV-755 record a grade search result when ILR does not exist

ben1stone:
Search results were previously found by finding the ILR record and then adding the certificate data to the results. However, Certificates could exist without an ILR record. I have added a condition where we map through the search results filling in the certificate data and an else so that if no results were found but a separate certificate does exist a search result would be created and added to the list.

I have also pulled out setting the basic and extra information into their own methods.


CodeWolf86 (changes requested):
1 Sonar code smell to deal with if we can, if we can rejig some of the conditionals to bring it in line ( e.g. short circuiting if the submitted log entry is null? )

And the standard question, do we have a unit test that will confirm a search result comes back with the completed certificate even if no ILR record exists?

CodeWolf86 (approved):

PR #1234, opened 2021-06-24
SV-658 save StandardReference and versions to the database

techtoniq:
Note the commit from 23rd July was me rebasing from SV-657


src/SFA.DAS.AssessorService.Application.Api/Validators/EpaOrganisationValidator.cs
```diff
@@ -513,8 +513,10 @@ public ValidationResponse ValidatorUpdateEpaOrganisationContactRequest(UpdateEpa
             RunValidationCheckAndAppendAnyError("OrganisationId", CheckIfOrganisationNotFound(request.OrganisationId), validationResult, ValidationStatusCode.NotFound);
             if (!validationResult.IsValid) return validationResult;
 
-            RunValidationCheckAndAppendAnyError("OrganisationId", CheckIfOrganisationStandardAlreadyExists(request.OrganisationId, request.StandardCode), validationResult, ValidationStatusCode.AlreadyExists);
-            if (!validationResult.IsValid) return validationResult;
+            //  SV-658 / SV-659 Now that we have versions, it is possible to register a version where the organisation+standard already exists.
+            //  So this validation is now redundant.
```
CodeWolf86, line 517: Should this validation be deleted as it's no longer required, or should it be upgraded to now throw an error if the standard versions being sent in already exist and are approved for?

src/SFA.DAS.AssessorService.Data/RegisterRepository.cs
```diff
@@ -77,29 +77,71 @@ public async Task<string>CreateEpaOrganisationStandard(EpaOrganisationStandard o
...
+
+                    foreach (var deliveryAreaId in deliveryAreas.Distinct())
                     {
-                        organisationStandard.OrganisationId, organisationStandard.StandardCode,
-                        organisationStandard.EffectiveFrom, organisationStandard.EffectiveTo,
-                        organisationStandard.DateStandardApprovedOnRegister, organisationStandard.Comments,
-                        organisationStandard.ContactId, organisationStandard.OrganisationStandardData
-                    }).Single();
+                        connection.Execute("INSERT INTO OrganisationStandardDeliveryArea ([OrganisationStandardId],DeliveryAreaId, Status) VALUES " +
+                                            "(@osdaId, @deliveryAreaId,'Live'); ",
```
CodeWolf86, line 106: Bug here, `@osdaId` doesn't match the object named parameter being passed in, it throws an exception

src/SFA.DAS.AssessorService.Data/RegisterRepository.cs
```diff
@@ -77,29 +77,71 @@ public async Task<string>CreateEpaOrganisationStandard(EpaOrganisationStandard o
...
-                    connection.Execute("INSERT INTO OrganisationStandardDeliveryArea ([OrganisationStandardId],DeliveryAreaId, Status) VALUES " + 
-                                        "(@osdaId, @deliveryAreaId,'Live'); ",
-                                    new { osdaId, deliveryAreaId}
-                                    );
-                }                      
-                        
-                return osdaId;
-            }
-            
+                    foreach (var version in organisationStandard.StandardVersions)
```
CodeWolf86, line 124: Like above, if it already exists it falls into an update, do we want to do a check here if the org standard version exists ( if we don't update the other validation rule called out in a comment ) otherwise we get a key constraint error.

I had an issue whereby this page errored, and i had to refresh then it kept falling over as it wouldn't get past the key constraint

CodeWolf86 (changes requested):

src/SFA.DAS.AssessorService.Application.Api/Validators/EpaOrganisationValidator.cs
CodeWolf86: After talking with Al, I think we should re-instate and improve the rule to be

If we're adding a new full standard, and it exists, and it's not withdrawn -> failure
if it is withdrawn -> success

if adding a version, and the version already exists and not withdrawn -> failure
if it is withdrawn -> success

Overall it's not great as once the error is returned, the admin person can't do anything about it but it at least maintains existing behaviour with the versioning upgrade.

CodeWolf86 (approved):

PR #1248, opened 2021-07-01
SV-912 API changes to support withdrawal applications page

techtoniq:


CodeWolf86 (approved):

src/SFA.DAS.AssessorService.Application/Handlers/Apply/Review/WithdrawalApplicationsHandler.cs
```diff
@@ -33,8 +37,40 @@ public async Task<PaginatedList<ApplicationSummaryItem>> Handle(WithdrawalApplic
...
 
+            if(items.Any())
+            {
+                // only want to do this once.
+                var allEnrolledVersions = await _standardRepository.GetEpaoRegisteredStandardVersions(items.First().EndPointAssessorOrganisationId);
+
+                // Get versions for each standard and determine the withdrawal application type.
+                foreach (var item in items)
+                {
+                    var allEnrolledStandardVersions = allEnrolledVersions.Where(v => v.IFateReferenceNumber == item.StandardReference).ToList();
```
CodeWolf86, line 48: Can we just tweak this method a bit?

The comments can probably be removed I think
// Only want to do this one  - is shown by the call outside the loop
// Get Versions for each standard  - seems to be in the wrong place? or outdated?

var allEnrolledStandardVersions - maybe enrolledVersionsForStandard?  the "All" bit I think confuses with the previous call

src/SFA.DAS.AssessorService.Application/Handlers/Apply/Review/WithdrawalApplicationsHandler.cs
```diff
@@ -33,8 +37,40 @@ public async Task<PaginatedList<ApplicationSummaryItem>> Handle(WithdrawalApplic
...
+                }
+            }
+
+
             return new PaginatedList<ApplicationSummaryItem>(items.ToList(),
                     organisationApplicationsResult.TotalCount, request.PageIndex, request.PageSize, request.PageSetSize);
         }
+
+        // SV-912 Helper to generate the type of withdrawal
+        private string GetWithdrawalApplicationType(ApplicationSummaryItem item, IEnumerable<StandardVersion> allEnrolledVersions)
```
CodeWolf86, line 59: Same here for tweaks
allEnrolledVersions -> enrolledVersionsForStandard?

then inside the loop
allVersions -> maybe versionNumberList?

I think the names are a hangover from the previous implementation but i think a few adjustments makes it more explicit.

CodeWolf86 (approved):

PR #1252, opened 2021-07-05
CON-3705 Limit the computed column size to help prevent an index which include…

ChrisJWoodcock:
…s it from exceeding the byte limit


VasanthaKasirajan3008 (dismissed):

kola-samuel (approved):
Approving  based on information that  rebuilt during a deployment  times out

PR #1307, opened 2021-07-30
SV-1122_ApprovalsExtract_and_Learner_table_and_PopulateLearner

alandburns:
Adds Learner table as destination for Approvals data in EPAO Certification, Pipeline and Opportunity Finder.


alandburns:
The Sonar Cloud check "Fail" is 100%  incorrect.
The DELETE should be without any WHERE clause, is within a Transaction and is intended to allow the transaction to rollback if the INSERT does not complete, hence leaving the table contents intact.

CodeWolf86 (approved):

PR #1309, opened 2021-07-30
SV-925 Approvals Standard Cache

CodeWolf86:
This PR is a knock on from the SV_925 Standards Cache update into approvals which included some tech debt.
- This changed version from decimal to string to proliferate through the service
- And has updated to store the new version major and minor fields from course api to aid in ordering in local logic in standards.


ben1stone (changes requested):

src/SFA.DAS.AssessorService.Application.UnitTests/Handlers/Standards/WhenHandlingGetCalculatedStandardVersionForApprenticeshipHandler.cs
```diff
@@ -58,7 +58,7 @@ public class WhenHandlingGetCalculatedStandardVersionForApprenticeshipHandler
 
             foreach (var s in standards)
             {
-                s.Version = baseVersion;
+                s.Version = baseVersion.ToString();
```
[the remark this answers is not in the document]
CodeWolf86: Yep that makes sense, have updated.

src/SFA.DAS.AssessorService.Data.IntegrationTests/StandardsGetTests.cs
```diff
@@ -32,14 +32,14 @@ public void SetupStandardsTable()
...
-        [TestCase("ST0001", "", "ST0001_1.2")]
-        [TestCase("ST0001", "1.10", "ST0001_1.10", Ignore = "Test will fail until version is stored as string (SV-568)")]
-        [TestCase("ST0001", "1.12", "ST0001_1.12", Ignore = "Test will fail until version is stored as string (SV-568)")]
+        //[TestCase("ST0001", "1", "ST0001_1.0")]
+        //[TestCase("ST0001", "1.0", "ST0001_1.0")]
+        //[TestCase("ST0001", "1.00", "ST0001_1.0")]
+        //[TestCase("ST0001", "1.1", "ST0001_1.1")]
+        //[TestCase("ST0001", "1.2", "ST0001_1.2")]
+        //[TestCase("ST0001", "", "ST0001_1.2")]
+        [TestCase("ST0001", "1.10", "ST0001_1.10")]
```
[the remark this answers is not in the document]
CodeWolf86: I've removed some of the test cases that I think make sense.

src/SFA.DAS.AssessorService.Application/Infrastructure/OuterApi/GetStandardByIdResponse.cs
```diff
@@ -25,7 +25,13 @@ public class StandardDetailResponse
         public int Level { get; set; }
 
         [JsonProperty("version")]
-        public decimal Version { get; set; }
+        public string Version { get; set; }
```
StuartReeve102, line 28: This refers to multiple classes, where you have added a string Version and interger VersionMajor and VersionMinor, properties, shouldn't the string Version property be a calculated property based on VersionMajor and VersionMinor?

src/SFA.DAS.AssessorService.Database/Tables/Standards.sql
```diff
@@ -4,7 +4,7 @@
 	[IFateReferenceNumber] VARCHAR(10) NOT NULL,
 	[LarsCode] INT NULL,
 	[Title] VARCHAR(1000) NOT NULL,
-	[Version] DECIMAL(18,1) NULL,
+	[Version] VARCHAR(20) NULL,
```
StuartReeve102, line 7: Do we need the Version column?

StuartReeve102 (changes requested):

src/SFA.DAS.AssessorService.Application/Infrastructure/OuterApi/GetStandardByIdResponse.cs
CodeWolf86: Given the data from Ifate comes in as a string, it seemed best to store that as is so we hold the original value. The VersionMajor VersionMinor is the calculated aspect from splitting the original field. The sole purpose of the split was purely for ordering purposes at the database level. It is a little bit duplicated but it seemed the simplest option.

The same applies here in the assessors space, only using the major and minor for ordering where version order is important while maintaining the original string value. It is probably more appropriate here for a calculated value but I was trying to limit the changes given how much of an impact just changing the type caused. In addition I knew we were coming back into the assessors service post approvals which would allow further tidy up once back in the space.

src/SFA.DAS.AssessorService.Database/Tables/Standards.sql
CodeWolf86: As above comment.

StuartReeve102 (approved):

ben1stone (approved):

ben1stone (approved):

PR #1325, opened 2021-09-09
CON-3878 Import Ilr when draft certificate exists

ChrisJWoodcock:


cofaulco (commented):
should there be a unit test where the IlrRecord is updated in the scenario when a draft certificate exists?

cofaulco (dismissed):

cofaulco (approved):

PR #1328, opened 2021-09-15
SV-1238 Approvals Extract API

techtoniq:


alandburns (approved):
Tested locally, and is working for me.

src/SFA.DAS.AssessorService.Application/Mapping/AutoMapperProfiles/ImportApprovalsProfile.cs
```diff
@@ -0,0 +1,14 @@
+﻿using AutoMapper;
+
+namespace SFA.DAS.AssessorService.Application.Mapping.AutoMapperProfiles
+{
+    public class ImportApprovalsProfile : Profile
+    {
+        public ImportApprovalsProfile()
+        {
+            CreateMap<Infrastructure.OuterApi.Learner, Domain.Entities.ApprovalsExtract>()
+                //.ForMember(dst => dst.ULN, opt => opt.MapFrom(src => src.ULN))
```
CodeWolf86, line 10: Can lose the comment for neatness

src/SFA.DAS.AssessorService.Application/Handlers/Approvals/ImportApprovalsHandler.cs
```diff
@@ -0,0 +1,120 @@
...
+            }
+
+            return Unit.Value;
+        }
+
+        private async Task<int> GetSettingAsInt(string name)
+        {
+            int value = 0;
+
+            var settingValue = await _settingRepository.GetSetting(name);
```
CodeWolf86, line 94: What was the thinking behind the _settingRepository and the database over the EmployerConfig that gets injected and adding a setting?

CodeWolf86 (changes requested):

src/SFA.DAS.AssessorService.Application/Handlers/Approvals/ImportApprovalsHandler.cs
techtoniq: It was a requirement in the JIRA / Wiki page - "Note, the Settings table, to hold name value pairs, in Assessor Db das-prd-ass-db is currently used by the ILR update processing. This will enable soft configuration of the batch size and time overlap for the /api/learners call - and will require a post-deployment script to configure at deployment time."   [](https://skillsfundingagency.atlassian.net/wiki/spaces/NDL/pages/3133210746/Approvals+database+-+SV#Assessor-sevice-API-to-Update-Approvals-Extract)

PR #1334, opened 2021-09-22
Sv 1242 include standard ref in cert data

anilkdhiman6:


alandburns (changes requested):
Line 147 is now superflous
certData.StandardReference = standardVersion.IfateReferenceNumber;
as StandardReference is set at line 116 and the StandardRefernce is the same for all versins of a standard

PR #1342, opened 2021-09-26
Add script to alter datatype of Standard Version - from decimal to va…

anilkdhiman6:
…rchar(20). Post database deployment


alandburns (changes requested):
This will not work.
The steps needs to be.
1. use pre-deployment to clear the version column
2. deployment (DACPAC) will run and make the Alter Column
3. post-deployment to recreate the version based on the RHS of StandardUId

PR #1351, opened 2021-10-07
SV-1218 Providers table to be used as lookup for Approvals in to Certification

techtoniq:


src/SFA.DAS.AssessorService.Data/ApprovalsExtractRepository.cs
```diff
@@ -139,5 +144,72 @@ public async Task<int> PopulateLearner()
...
+            }
+        }
+
+        private async Task<IEnumerable<int>> UkprnsInExtractNotInProviders()
+        {
+            var ukprns = await _unitOfWork.Connection.QueryAsync<int>("SELECT DISTINCT Ukprn FROM ApprovalsExtract WHERE Ukprn NOT IN (SELECT DISTINCT Ukprn FROM Providers)");
+            return ukprns;
+        }
+
+        private async Task<IEnumerable<int>> UkprnsInLearnersNotInProviders()
```
CodeWolf86, line 192: Think we can lose this method now?

PR #1361, opened 2021-10-19
SV-1277 fix integration tests

ben1stone:
I've added a Delete CertificateLogs method as Certificates could not be deleted in tear down. I've also made a few tweaks to the test data to avoid clashes.


techtoniq (dismissed):

alandburns (dismissed):
Re-based PR to allow the fixed integration tests to be used for the StandardCollation removal.

PR #1376, opened 2021-11-15
SV-1305 Expired Financial Stage 1

techtoniq:


alandburns (changes requested):
If I am reading this right, it is not taking account of when the Organisation has been marked as being Exempt from Financial review and hence will not be required to (re)do a financial review, but this does not essentially mean that the financial due date does not have an out of date value.
You cannot rely solely on a null "FinancialDueDate" value.

techtoniq:
Per discussion with Al - DO NOT MERGE. Awaiting confirmation on the business process/comms required where the data does not have a value for the "FinancialExempt" flag.

CodeWolf86 (approved):

src/SFA.DAS.AssessorService.Application/Helpers/FinancialReviewStatusHelper.cs
```diff
@@ -0,0 +1,36 @@
+﻿using SFA.DAS.AssessorService.Api.Types.Models.AO;
+using SFA.DAS.AssessorService.Application.Interfaces;
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace SFA.DAS.AssessorService.Application.Helpers
+{
+    public static class FinancialReviewStatusHelper
```
CodeWolf86, line 9: Could we get some unit tests around confirming the behaviour of when you are financially exempt?

CodeWolf86 (changes requested):

CodeWolf86 (approved):

PR #1379, opened 2021-11-23
SV-1281 manual add api without SV-1290

ben1stone:
Fresh PR before SV-1290 tech debt work

### DIFF
--- a/src/SFA.DAS.AssessorService.Api.Types/Models/AO/EpaOrganisationStandard.cs
+++ b/src/SFA.DAS.AssessorService.Api.Types/Models/AO/EpaOrganisationStandard.cs
@@ -9,7 +9,7 @@ namespace SFA.DAS.AssessorService.Api.Types.Models.AO
         public string OrganisationId { get; set; }
         public int StandardCode { get; set; }
         public string StandardReference { get; set; }
-        public List<string> StandardVersions { get; set; }
+        public List<OrganisationStandardVersion> StandardVersions { get; set; }
         public DateTime? EffectiveFrom { get; set; }
         public DateTime? EffectiveTo { get; set; }
         public DateTime? DateStandardApprovedOnRegister { get; set; }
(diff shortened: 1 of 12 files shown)


src/SFA.DAS.AssessorService.Data/RegisterRepository.cs
```diff
@@ -101,21 +101,21 @@ public async Task<string> CreateEpaOrganisationStandard(EpaOrganisationStandard
 
             if (null != organisationStandard.StandardVersions)
             {
-                foreach (var version in organisationStandard.StandardVersions)
+                foreach (var standardVersion in organisationStandard.StandardVersions)
                 {
-                    var standardUid = $"{organisationStandard.StandardReference.Trim()}_{version.Trim()}";
+                    var standardUid = $"{organisationStandard.StandardReference.Trim()}_{standardVersion.Version.Trim()}";
```
CodeWolf86, line 106: Now we've changed the input from string Versions to the actual Organisation Standard Version object type. Is there any reason we can't supply the StandardUId rather than crafting it?

src/SFA.DAS.AssessorService.Api.Types/Models/AO/EpaOrganisationStandard.cs
```diff
@@ -9,7 +9,7 @@ public class EpaOrganisationStandard
         public string OrganisationId { get; set; }
         public int StandardCode { get; set; }
         public string StandardReference { get; set; }
-        public List<string> StandardVersions { get; set; }
+        public List<OrganisationStandardVersion> StandardVersions { get; set; }
```
CodeWolf86, line 12: This has changed due to the request, I can't see the same change on the model Admin Service side, should there be? (Admin side screenshot)
image

CodeWolf86 (changes requested):

CodeWolf86:
Slight tweak from the Sonar Cloud issue as well.https://sonarcloud.io/project/issues?id=SkillsFundingAgency_das-assessor-service&pullRequest=1379&resolved=false&types=CODE_SMELL

ben1stone:
Closed. A new design based on merging instead of manually adding will be implemented instead

PR #1394, opened 2021-12-21
SV-1350-SQL Tweaks for StaffSearch

CodeWolf86:


cofaulco (approved):

ChrisJWoodcock (changes requested):
The new index on the certificates table only includes the computed fields - I think that it should be containing the primary key as well so that the index can be used to refer to the other data?

ChrisJWoodcock (changes requested):
There is no index on the new computed column in the learner table - so that would not make a different to the performance

CodeWolf86:
> The new index on the certificates table only includes the computed fields - I think that it should be containing the primary key as well so that the index can be used to refer to the other data?

I was working on the basis that at minimum, if the Index has the same fields in it as the where clause it can make use of it as part of the query. But I take your point that having the ID in there as well can make a difference when connecting up to various related data.

ChrisJWoodcock (dismissed):

ChrisJWoodcock (approved):

PR #1396, opened 2021-12-30
Con-4248 get employer details v2

kola-samuel:


ChrisJWoodcock (approved):
I've updated the base branch CON-4164 to match the column name used here.

cofaulco (approved):

PR #1397, opened 2022-01-04
SV-1322 Organisation Merge

techtoniq:


src/SFA.DAS.AssessorService.Application.Api/Services/OrganisationMergingService.cs
```diff
@@ -0,0 +1,280 @@
...
+                Status = MergeOrganisationStatus.Approved,  // auto-approved for now
+            };
+            _dbContext.MergeOrganisations.Add(mo);
+
+            return mo;
+        }
+
+        private void CreateStandardsSnapshot(MergeOrganisation mo, string endpointAssessorOrganisationId, string replicates)
+        {
+            var organisationStandards = _dbContext.OrganisationStandard.Where(e => e.EndPointAssessorOrganisationId == endpointAssessorOrganisationId);
```
CodeWolf86, line 106: Can improve this by using `Includes` on the EF call:

```
          var organisationStandards = _dbContext.OrganisationStandard.
                Include(a => a.OrganisationStandardVersions).
                Where(e => e.EndPointAssessorOrganisationId == endpointAssessorOrganisationId);

foreach(var osv in os.OrganisationStandardVersions)
                {

                }
```

src/SFA.DAS.AssessorService.Application.Api/Services/OrganisationMergingService.cs
```diff
@@ -0,0 +1,280 @@
...
+            foreach (var os in organisationStandards)
+            {
+                mo.MergeOrganisationStandards.Add(CreateMergeOrganisationStandard(replicates, os));
+
+                var organisationStandardVersions = _dbContext.OrganisationStandardVersion.Where(e => e.OrganisationStandardId == os.Id);
+                foreach (var osv in organisationStandardVersions)
```
CodeWolf86, line 112: Which if you tweak the above means you can streamline this to:
`os.OrganisationStandardVersions.Select(x => CreateMergeOrganisationStandardVersion(replicates, x));`

And similiar to delivery areas

src/SFA.DAS.AssessorService.Application.Api/Services/OrganisationMergingService.cs
```diff
@@ -0,0 +1,280 @@
...
+                    mo.MergeOrganisationStandardDeliveryAreas.Add(CreateMergeOrganisationStandardDeliveryArea(replicates, osda));
+                }
+            }
+        }
+
+        private MergeOrganisationStandard CreateMergeOrganisationStandard(string replicates, OrganisationStandard sourceOrganisationStandard)
+        {
+            var mos = new MergeOrganisationStandard()
+            {
+                EndPointAssessorOrganisationId = sourceOrganisationStandard.EndPointAssessorOrganisationId,
```
CodeWolf86, line 129: Part of me does and doesn't like this. Part of me wants to make use of a Mapper, a constructor to encapsulate into the class itself, or to do some form of implicit cast so that the logic is moved out of the service and into the respective objects. But this at least does show you what it's doing so something to talk about maybe?

src/SFA.DAS.AssessorService.Application.Api/Services/OrganisationMergingService.cs
```diff
@@ -0,0 +1,280 @@
...
+
+            // Grab the primary contact for the primary organisation because we're going to need it.
+            var primaryContact = _dbContext.Contacts.FirstOrDefault(c => c.EndPointAssessorOrganisationId == primaryOrganisation.EndPointAssessorOrganisationId && c.Email == primaryOrganisation.PrimaryContact);
+
+            // Read all the standards from the secondary organisation.
+
+            foreach(var secondaryOrganisationStandard in secondaryOrganisation.OrganisationStandards)
+            {
+                // Does the primary organisation have this standard already?
+                var primaryOrganisationStandard = primaryOrganisation.OrganisationStandards.FirstOrDefault(pos => pos.StandardCode == secondaryOrganisationStandard.StandardCode);
```
CodeWolf86, line 188: If there is no primary contact, should we throw again?

src/SFA.DAS.AssessorService.Application.Api/Services/OrganisationMergingService.cs
```diff
@@ -0,0 +1,280 @@
...
+
+            MergeOrganisationStandardsAndVersions(
+                primaryOrganisation,
+                secondaryOrganisation,
+                mergedByUserId,
+                secondaryStandardsEffectiveTo);
+
+            // Create the "After" snapshot.
+
+            CreateStandardsSnapshot(mergeOrganisation, primaryOrganisation.EndPointAssessorOrganisationId, "After");
```
CodeWolf86, line 61: Between the Primary and Secondary snapshot, we're somewhat "abusing" EF, the first snapshot creates the copy, the merge happens against all the objects tracked against the EF context, then when the secondary snapshot comes along, we're relying on it returning to us the objects in the context that have yet to be committed. Allowing us to create what the secondary snapshot will look like even before it's saved.

I don't think there is a problem with this, I'm just noting it here for discussion

src/SFA.DAS.AssessorService.Application.Api/Services/OrganisationMergingService.cs
```diff
@@ -0,0 +1,280 @@
...
+        private void MergeOrganisationStandardsAndVersions(Organisation primaryOrganisation, Organisation secondaryOrganisation, Guid createdByUserId, DateTime secondaryStandardsEffectiveTo)
+        {
+            // @ToDo: a bit monolithic - can this be refactored in to something more readable?
+
+            // Grab the primary contact for the primary organisation because we're going to need it.
+            var primaryContact = _dbContext.Contacts.FirstOrDefault(c => c.EndPointAssessorOrganisationId == primaryOrganisation.EndPointAssessorOrganisationId && c.Email == primaryOrganisation.PrimaryContact);
+
+            // Read all the standards from the secondary organisation.
+
+            foreach(var secondaryOrganisationStandard in secondaryOrganisation.OrganisationStandards)
```
CodeWolf86, line 185: You've used the lazy loading here as opposed to earlier on the snapshots 👍

src/SFA.DAS.AssessorService.Application.Api/Services/OrganisationMergingService.cs
```diff
@@ -0,0 +1,280 @@
...
+                // Does the primary organisation have this standard already?
+                var primaryOrganisationStandard = primaryOrganisation.OrganisationStandards.FirstOrDefault(pos => pos.StandardCode == secondaryOrganisationStandard.StandardCode);
+
+                if (null == primaryOrganisationStandard)
+                {
+                    // No - so add the standard to the primary organisation.
+
+                    primaryOrganisationStandard = new OrganisationStandard()
+                    {
+                        DateStandardApprovedOnRegister = secondaryOrganisationStandard.DateStandardApprovedOnRegister,
```
CodeWolf86, line 196: Should this date be the date the secondary was approved, or "now" that's when the primary will be enabled from the merge date?

src/SFA.DAS.AssessorService.Application.Api/Services/OrganisationMergingService.cs
```diff
@@ -0,0 +1,280 @@
...
+                        OrganisationStandardData = secondaryOrganisationStandard.OrganisationStandardData,
+                        Comments = $"{secondaryOrganisationStandard.Comments} ** this standard has been merged from Organisation {secondaryOrganisation.EndPointAssessorOrganisationId}",
+                        ContactId = primaryContact?.Id,
```
CodeWolf86, line 199: I'm also back to whether you would want to put some sort of `Clone` method on the OrganisationStandard entity to allow you to abstract the logic out of this merge service, then can test against that specifically via unit tests to break this down. Appreciate we have to do some form of copy as we don't want to copy across the EF context tracking.

src/SFA.DAS.AssessorService.Application.Api/Services/OrganisationMergingService.cs
```diff
@@ -0,0 +1,280 @@
...
+                        Status = secondaryOrganisationStandard.Status,
+
+                        OrganisationStandardVersions = new List<OrganisationStandardVersion>(),
+                        OrganisationStandardDeliveryAreas = new List<OrganisationStandardDeliveryArea>(),
+                    };
+                    primaryOrganisation.OrganisationStandards.Add(primaryOrganisationStandard);
+                }
+
+                // Now read all the versions for this standard for the secondary organisation.
+                var secondaryOrganisationStandardVersions = _dbContext.OrganisationStandardVersion.Where(sosv => sosv.OrganisationStandardId == secondaryOrganisationStandard.Id);
```
CodeWolf86, line 214: Can you do `secondaryOrganisationStandard.OrganisationStandardVersions` ?

src/SFA.DAS.AssessorService.Application.Api/Services/OrganisationMergingService.cs
```diff
@@ -0,0 +1,280 @@
...
+                foreach(var secondaryOrganisationStandardVersion in secondaryOrganisationStandardVersions)
+                {
+                    // Does the standard version exist for this standard in the primary organisation?
+
+                    var primaryOrganisationStandardVersion = primaryOrganisationStandard.OrganisationStandardVersions.FirstOrDefault(posv => posv.StandardUId == secondaryOrganisationStandardVersion.StandardUId && posv.Version == secondaryOrganisationStandardVersion.Version);
```
CodeWolf86, line 219: StandardUID should be enough, it's a concatenation of IFAte ref + version -> ST0016_1.0

CodeWolf86 (changes requested):
Some suggestions from the draft so far

src/SFA.DAS.AssessorService.Data/AssessorDbContext.cs
```diff
@@ -114,7 +121,7 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
             
             modelBuilder.Entity<OrganisationStandardVersion>()
                 .ToTable("OrganisationStandardVersion")
-                .HasKey(c => new { c.OrganisationStandardId, c.StandardUId });
+                .HasKey(c => new { c.OrganisationStandardId, c.StandardUId, c.Version });
```
CodeWolf86, line 124: Can drop the version here I think as well for primary key

CodeWolf86 (changes requested):

CodeWolf86 (changes requested):

src/SFA.DAS.AssessorService.Application.Api/Services/OrganisationMergingService.cs
```diff
@@ -0,0 +1,314 @@
...
+    {
+        public const string InProgress = "In Progress";
+        public const string Approved = "Approved";
+        public const string Completed = "Completed";
+        public const string Reverted = "Reverted";
+    }
+
+    public static class ReplicationType
+    {
+        public const string Before = "Before";
```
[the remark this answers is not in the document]
techtoniq: Nice, I like that. Like you say, tricky for In Progress

src/SFA.DAS.AssessorService.Application.Api/Services/OrganisationMergingService.cs
```diff
@@ -0,0 +1,314 @@
...
+        {
+            _dbContext = dbContext;
+        }
+
+        public async Task<MergeOrganisation> MergeOrganisations(Organisation primaryOrganisation, Organisation secondaryOrganisation, DateTime secondaryStandardsEffectiveTo, string actionedByUser)
+        {
+            if (null == primaryOrganisation) throw new ArgumentNullException(nameof(primaryOrganisation), "primaryOrganisation must be specified.");
+            if (null == secondaryOrganisation) throw new ArgumentNullException(nameof(secondaryOrganisation), "secondaryOrganisation must be specified.");
+
+            var executionStrategy = _dbContext.Database.CreateExecutionStrategy();
```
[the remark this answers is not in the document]
techtoniq: I added this because of a runtime exception where you can't have explicit transactions on the default execution strategy.

src/SFA.DAS.AssessorService.Application.Api/Services/OrganisationMergingService.cs
```diff
@@ -0,0 +1,314 @@
...
+            };
+            _dbContext.MergeOrganisations.Add(mo);
+
+            return mo;
+        }
+
+        private void CreateStandardsSnapshot(MergeOrganisation mo, Organisation organisation, string replicates)
+        {
+            // Make sure the standards are loaded.
+            _dbContext.Entry(organisation)
```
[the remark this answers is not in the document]
techtoniq: In testing I was seeing that the object graphs were not being loaded.

alandburns (approved):

PR #1398, opened 2022-01-05
Sv 889 handle reapplies for withdrawn standard assessor svc

anilkdhiman6:


src/SFA.DAS.AssessorService.Web/Controllers/Apply/StandardController.cs
```diff
@@ -353,6 +381,35 @@ private IEnumerable<StandardVersionApplication> ApplyVersionStatuses(IEnumerable
                 }
             }
 
+            var withdrawals = previousWithdrawals.Where(x => x.StandardApplicationType == StandardApplicationTypes.VersionWithdrawal);
```
CodeWolf86, line 384: I think there are some holes in the logic added to this method:
- It doesn't take into account the current version status, if a version is approved, and it previously had a withdrawal application, it's going to get overridden, in what cases should the withdrawal status be set? (ie not approved or in progress)
- You would only want the latest withdrawal applications for the versions of the standard, not all. ( ie withdraw, apply withdraw again )
- EPA changed could happen across multiple versions not just the latest so needs to be taken into account.

I think you would be better off altering the `MapUnapprovedVersionStatus` method, whereby you are looping through the versions and their state, pass in the previous withdrawal applications into the method, and then under the withdrawn status, apply a shorter version of the logic
- For the version that's withdrawn, find it's withdrawal application and it's withdrawal date.
- Change it's version status from Withdrawn to the opt in or 7 apply depending.

or something similiar to the above?

CodeWolf86 (changes requested):
Few changes to the code submitted, Also needs some unit tests around the opt in with the opt in following withdrawal. Along with changes to existing unit tests around the StandardController with the determining of the version status as that's a critical part of the change.

src/SFA.DAS.AssessorService.Application.Api.Client/Clients/ApplicationApiClient.cs
```diff
@@ -73,6 +73,44 @@ public async Task<ApplicationResponse> GetApplicationForUser(Guid id, Guid userI
                 return await RequestAndDeserialiseAsync<ApplicationResponse>(request, $"Could not retrieve application {id} for user {userId}");
             }
         }
+        public async Task<ApplicationResponse> GetLastWithdrawnApplication(Guid orgId, int? standardCode)
```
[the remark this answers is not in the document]
anilkdhiman6: Removed

src/SFA.DAS.AssessorService.Application.Api/Controllers/Apply/ApplicationController.cs
```diff
@@ -82,6 +82,25 @@ public async Task<ActionResult<ApplicationResponse>> GetApplication(string id)
             return Ok(await _mediator.Send(new GetApplicationRequest(Guid.Parse(id))));
         }
 
+
+        [HttpGet("{orgId}/application/withdrawn/{standardCode}", Name = "GetAllWithdrawnApplicationsForStandard")]
+        [SwaggerResponse((int)HttpStatusCode.OK, Type = typeof(ApplicationResponse))]
+        [SwaggerResponse((int)HttpStatusCode.InternalServerError, Type = typeof(ApiResponse))]
+        public async Task<ActionResult<List<ApplicationResponse>>> GetAllWithdrawnApplicationsForStandard(string orgId, string standardCode)
```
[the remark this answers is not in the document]
anilkdhiman6: Corrected

src/SFA.DAS.AssessorService.Application/Handlers/OrganisationStandards/OrganisationStandardVersionOptInHandler.cs
```diff
@@ -62,11 +62,19 @@ public async Task<OrganisationStandardVersion> Handle(OrganisationStandardVersio
                     Status = request.Status
                 };
 
-                var existingVersion = await _repository.GetOrganisationStandardVersionByOrganisationStandardIdAndVersion(orgStandard.Id, request.Version);
-                if (existingVersion != null)
-                    throw new InvalidOperationException("OrganisationStandardVersion already exists");
+                if (!request.OptInFollowingWithdrawal)
```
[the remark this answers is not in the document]
anilkdhiman6: amended as requested

src/SFA.DAS.AssessorService.Web/Controllers/Apply/StandardController.cs
```diff
@@ -121,7 +124,7 @@ public async Task<IActionResult> ConfirmStandard(Guid id, string standardReferen
                 // existing approved versions for this standard
                 var model = new StandardVersionApplicationViewModel { Id = id, StandardReference = standardReference };
                 model.SelectedStandard = new StandardVersionApplication(latestStandard);
-                model.Results = ApplyVersionStatuses(standardVersions).OrderByDescending(x => x.Version).ToList();
+                model.Results = ApplyVersionStatuses(standardVersions, allPreviousWithdrawalsForStandard, previousApplication).OrderByDescending(x => x.Version).ToList();
```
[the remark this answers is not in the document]
anilkdhiman6: Correct - this has been removed

src/SFA.DAS.AssessorService.Web/Controllers/Apply/StandardController.cs
```diff
@@ -311,7 +317,29 @@ private async Task<string> ApplicationStandardStatus(ApplicationResponse applica
...
+                return false;
+            }
+            else
+            {
+                return true;
+            }
+        }
+
+
+        private IEnumerable<StandardVersionApplication> ApplyVersionStatuses(IEnumerable<AppliedStandardVersion> versions,
```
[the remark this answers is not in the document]
anilkdhiman6: Removed default

src/SFA.DAS.AssessorService.Web/Controllers/Apply/StandardController.cs
```diff
@@ -311,7 +317,29 @@ private async Task<string> ApplicationStandardStatus(ApplicationResponse applica
...
+            if (previousWithdrawalDate > DateTime.UtcNow.AddMonths(-12) && prevApplyViaOptIn == true)
+            {
+                return true;
+            }
+            else if (previousWithdrawalDate < DateTime.UtcNow.AddMonths(-12) && !prevApplyViaOptIn == false)
+            {
+                return true;
+            }
+            else if (previousWithdrawalDate < DateTime.UtcNow.AddMonths(-12) && prevApplyViaOptIn == true)
+            {
```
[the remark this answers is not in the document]
anilkdhiman6: Updated - simplified

src/SFA.DAS.AssessorService.Web/Controllers/Apply/StandardController.cs
anilkdhiman6: Refactored - added `!= approved` status check

src/SFA.DAS.AssessorService.Web/Controllers/Apply/StandardController.cs
```diff
@@ -256,7 +259,10 @@ public async Task<IActionResult> OptInPost(Guid id, string standardReference, st
             var standards = await _standardVersionApiClient.GetStandardVersionsByIFateReferenceNumber(standardReference);
             var stdVersion = standards.First(x => x.Version.Equals(version, StringComparison.InvariantCultureIgnoreCase));
 
-            await _orgApiClient.OrganisationStandardVersionOptIn(id, contact.Id, org.OrganisationId, standardReference, version, stdVersion.StandardUId, $"Opted in by EPAO by {contact.Username}");
+            var previousApplication = await _applicationApiClient.GetPreviousApplication(application.OrganisationId, standardReference);
```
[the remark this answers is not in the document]
anilkdhiman6: Refactored - added additional logic to account for versions and correct sequence of events

src/SFA.DAS.AssessorService.Application.UnitTests/Handlers/Applications/GetApplicationHandlerTests/When_organisation_previous_application_is_requested.cs
```diff
@@ -0,0 +1,42 @@
...
+
+            ApplyRepository = new Mock<IApplyRepository>();
+            Handler = new GetPreviousApplicationsHandler(ApplyRepository.Object);
+        }
+
+
+        [Test]
+        public async Task Then_organisation_previous_application_is_retrieved()
+        {
+            await Handler.Handle(new GetPreviousApplicationsRequest(new Guid(), ""), new CancellationToken());
```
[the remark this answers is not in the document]
anilkdhiman6: Updated

src/SFA.DAS.AssessorService.Application.UnitTests/Handlers/Applications/GetApplicationHandlerTests/When_organisation_all_withdrawn_applications_are_requested.cs
```diff
@@ -0,0 +1,42 @@
...
+            ApplyRepository = new Mock<IApplyRepository>();
+            Handler = new GetAllWithdrawnApplicationsForStandardHandler(ApplyRepository.Object);
+        }
+
+
+        [Test]
+        public async Task Then_organisation_withdrawn_applications_are_retrieved()
+        {
+            await Handler.Handle(new GetAllWithdrawnApplicationsForStandardRequest(new Guid(), 59), new CancellationToken());
+            ApplyRepository.Verify(r => r.GetAllWithdrawnApplicationsForStandard(It.IsAny<Guid>(),It.IsAny<int>()), Times.Once);
```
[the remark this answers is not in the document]
anilkdhiman6: Updated

src/SFA.DAS.AssessorService.Application/Handlers/EpaOrganisationHandlers/CreateEpaOrganisationStandardHandler.cs
```diff
@@ -59,7 +59,14 @@ public async Task<string> Handle(CreateEpaOrganisationStandardRequest request, C
 
             var organisationStandard = MapOrganisationStandardRequestToOrganisationStandard(request);
            
-            return await _registerRepository.CreateEpaOrganisationStandard(organisationStandard, request.DeliveryAreas);
+            if (request.ApplyFollowingWithdrawal)
```
CodeWolf86, line 62: So if it's an apply following a withdrawal, we update the records which is great. If it's not then we add the records which is also great.

One scenario I thought of which i'm unsure if it's a valid scenario is, can the application following a withdrawal be for many versions? If it can, then you could be in a scenario where you withdrew 1.0 but never 1.1 as that's what you're applying for now.

In that scenario you need to update 1.0 and add 1.1.

If you withdrew from a version, and still hold the standard, then you can only opt in / short apply for one version, so the above scenario is invalid in this case.

If you've withdrawn from the entire standard, then choose to apply again, can you choose to apply for more than one version? in which case the current logic would fail.

src/SFA.DAS.AssessorService.Application/Handlers/OrganisationStandards/OrganisationStandardVersionOptInHandler.cs
```diff
@@ -61,12 +61,20 @@ public async Task<OrganisationStandardVersion> Handle(OrganisationStandardVersio
                     Comments = request.Comments,
                     Status = request.Status
                 };
+                
+                if (request.OptInFollowingWithdrawal)
+                {
+                    await _repository.UpdateOrganisationStandardVersion(entity);
```
CodeWolf86, line 67: Do you still need the existing version check in this branch of the code as well as a validation check, or are we happy that if it doesn't exist, then the update will just do nothing?

src/SFA.DAS.AssessorService.Web/Controllers/Apply/StandardController.cs
```diff
@@ -353,6 +386,35 @@ private IEnumerable<StandardVersionApplication> ApplyVersionStatuses(IEnumerable
                 }
             }
 
+            var withdrawals = previousWithdrawals.Where(x => x.StandardApplicationType == StandardApplicationTypes.VersionWithdrawal
```
CodeWolf86, line 389: This isn't what I meant when checking for Approved in the previous comment, You want approved withdrawal applications don't you?

CodeWolf86 (changes requested):

src/SFA.DAS.AssessorService.Application/Handlers/OrganisationStandards/OrganisationStandardVersionOptInHandler.cs
anilkdhiman6: Added existing version check in for completeness

src/SFA.DAS.AssessorService.Web/Controllers/Apply/StandardController.cs
anilkdhiman6: Yes - this has been corrected

src/SFA.DAS.AssessorService.Application/Handlers/EpaOrganisationHandlers/CreateEpaOrganisationStandardHandler.cs
anilkdhiman6: The behaviour is that :
1. System will detect that it has previously been withdrawn
2. System will update the versions but WILL NOT create the entries for the new additional version selected

anilkdhiman6: Modified SQL in repository- it will check whether the standard version(s) passed into the query exists - if it does not then it will create it

src/SFA.DAS.AssessorService.Web/Controllers/Apply/StandardController.cs
anilkdhiman6: `!= approved` has been corrected

Have tested this scenario and seems as if it works ok - unless theres a scenario I've missed,
I will catch up with Uma to see what hes getting in the test env

src/SFA.DAS.AssessorService.Data/RegisterRepository.cs
```diff
@@ -159,18 +159,43 @@ public async Task<string> CreateEpaOrganisationStandard(EpaOrganisationStandard
...
+            {
+                foreach (var version in orgStandard.StandardVersions)
                 {
-                    orgStandard.EffectiveFrom,
-                    orgStandard.EffectiveTo,
-                    orgStandard.Id
-                });
+                    var standardUid = $"{orgStandard.StandardReference.Trim()}_{version.Trim()}";
+
+                    await _unitOfWork.Connection.ExecuteAsync(
```
CodeWolf86, line 169: Can this be tweaked a bit? 
The code below detects for each individual org standard version, and then inserts if it doesn't exist, but while you're doing the insert, you could insert the correct effective from and effective to meaning you wouldn't need to update it afterwards?

You almost want to query for all versions of the OrgStandard First, then based on what exists and what doesn't, then update / create.

Alternatively, can you change the SQL so the IF NOT EXISTS has an else?
- IF NOT EXISTS
- INSERT INTO
- ELSE UPDATE

What do you prefer?

CodeWolf86 (changes requested):

src/SFA.DAS.AssessorService.Data/RegisterRepository.cs
anilkdhiman6: Agreed - added IF ELSE for SQL Query

```
IF NOT EXISTS
    INSERT
ELSE
   UPDATE
```

CodeWolf86 (dismissed):

src/SFA.DAS.AssessorService.Application/Handlers/OrganisationStandards/OrganisationStandardVersionOptInHandler.cs
```diff
@@ -63,10 +63,20 @@ public async Task<OrganisationStandardVersion> Handle(OrganisationStandardVersio
                 };
 
                 var existingVersion = await _repository.GetOrganisationStandardVersionByOrganisationStandardIdAndVersion(orgStandard.Id, request.Version);
-                if (existingVersion != null)
-                    throw new InvalidOperationException("OrganisationStandardVersion already exists");
 
-                await _repository.CreateOrganisationStandardVersion(entity);
+                if (request.OptInFollowingWithdrawal)
```
CodeWolf86, line 67:
```
                if (request.OptInFollowingWithdrawal && existingVersion != null)
                {
                    await _repository.UpdateOrganisationStandardVersion(entity);
                }

                else if(existingVersion != null)
                {
                        throw new InvalidOperationException("OrganisationStandardVersion already exists");
                }

               else
               {
                      await _repository.CreateOrganisationStandardVersion(entity);
               }
```


Think the logic needs tweaking as this could mean some bits fall through the gaps?

- if existing version not null and following withdrawal, we can update
- if existing version not null but not following a withdrawal, we throw
- if existing version is null, we can create.


in the existing logic, you could set opt in following withdrawal to true, but existing version could be null and nothing happens. Removing nesting is better for readability

src/SFA.DAS.AssessorService.Database/Tables/Apply.sql
```diff
@@ -16,7 +16,8 @@ CREATE TABLE [Apply](
 	[DeletedBy] [nvarchar](256) NULL,
     [StandardApplicationType]  NVARCHAR(60) NULL ,  
     [StandardReference] NVARCHAR(10) NULL ,    
- CONSTRAINT [PK_Apply] PRIMARY KEY CLUSTERED 
+	[ApplyViaOptIn] BIT NULL DEFAULT 0,
```
CodeWolf86, line 19: If we've added a new field, and this is being used to determine if they applied via opt in to help with the withdrawal logic when re-applying, you need to factor in how you're going to update this for live data that has already had applications that have opted in via apply.

CodeWolf86 (changes requested):

src/SFA.DAS.AssessorService.Database/Tables/Apply.sql
anilkdhiman6: Post deployment script added

src/SFA.DAS.AssessorService.Application/Handlers/OrganisationStandards/OrganisationStandardVersionOptInHandler.cs
anilkdhiman6: Agreed - Tidied Up / Updated as requested

CodeWolf86 (dismissed):

CodeWolf86 (approved):

alandburns (approved):

PR #1404, opened 2022-01-25
Sv 1305 expired stage1

anilkdhiman6:


CodeWolf86 (approved):

src/SFA.DAS.AssessorService.Web/Controllers/OrganisationStandardsController.cs
```diff
@@ -57,18 +57,21 @@ public async Task<IActionResult> Index(int? pageIndex)
...
+
+                    if ((!orgDataFinancialExempt.HasValue) || (orgDataFinancialExempt.Value == false))
                     {
-                        var financialDueDate = organisation.OrganisationData?.FHADetails?.FinancialDueDate;
-                        if (null == financialDueDate || (financialDueDate.Value.Date < DateTime.UtcNow.Date))
+                        if ((!orgDataFinancialDueDate.HasValue) || (orgDataFinancialDueDate.Value.Date < DateTime.UtcNow.Date))
                         {
-                            model.FinancialInfoStage1Expired = true;
-                            model.FinancialAssessmentUrl = this.Url.Action("StartOrResumeApplication", "Application");
+                            if (orgTypeFinancialExempt != ApplyTypes.FinancialReviewStatus.Exempt)
```
alandburns, line 70: It would be more efficient to check if the organisationtype is exempt, before looking at the organisation data, as these values would only be set for an organisation that is not exempt.

CodeWolf86: Following on from Als comment, I think this can actually be simplified. the `organisation.FinancialReviewStatus` that's returned from the Api, uses the `FinancialReviewStatusHelper.IsFinancialExempt` method to determine the status and then populate either `FinancialExempt` or `Required` so just checking that field alone should be enough ( as it was before the change in this PR ) as at the moment you are re-doing the check that's done in `FinancialReviewStatusHelper` for no real gain.

src/SFA.DAS.AssessorService.Web/Controllers/Apply/ApplicationController.cs
```diff
@@ -1275,5 +1309,30 @@ private static void UpdateValidationDetailsForAddress(Page page, List<Validation
...
 
         }
+
+        public async Task<FinancialExpiredViewModel> IsFinancialExpired(string epaoId, string redirAction, string redirController)
+        {
+            var model = new FinancialExpiredViewModel();
+            var organisation = await _orgApiClient.GetEpaOrganisation(epaoId);
+            if (organisation != null)
+            {
+                var orgDataFinancialExempt = organisation.OrganisationData?.FHADetails?.FinancialExempt;
```
CodeWolf86, line 1319: I think this here can be improved, the `GetEpaOrganisation` call returns the `organisation.FinancialReviewStatus` of Exempt or required which is done Api side with the same tests.

Can this not be:

```
                    if (organisation.FinancialReviewStatus != ApplyTypes.FinancialReviewStatus.Exempt)
                    {
                        model.FinancialInfoStage1Expired = true;
                        model.FinancialAssessmentUrl = this.Url.Action(redirAction, redirController);
                    }
```
?

Also this method doesn't need to return a financial assessment url , both the two instances that call this method only care about `FinancialInfoStage1Expired` being true or false, so the method could be reduced to only return this value. And when this is taken into account, the actual helper method could be removed entirely and where it's called be replaced with
```
var organisation = await _orgApiClient.GetEpaOrganisation(epaoId);
var financialExpired = organisation?.FinancialReviewStatus == ApplyTypes.FinancialReviewStatus.Exempt ?? false;
```

As in the areas that use the boolean, only if it's true does the `FinancialExpiredModel` come into play. so for example:
```
            var financialExpired = await IsFinancialExpired(epaoid, "StartOrResumeApplication", "Application");

            ApplicationResponseViewModel applicationsResponseVm = new ApplicationResponseViewModel();
            applicationsResponseVm.ApplicationResponse = existingApplications;
            applicationsResponseVm.FinancialInfoStage1Expired = financialExpired.FinancialInfoStage1Expired;
            applicationsResponseVm.FinancialAssessmentUrl = financialExpired.FinancialAssessmentUrl;
```
Becomes
```
            var organisation = await _orgApiClient.GetEpaOrganisation(epaoId);
            var financialExpired = organisation?.FinancialReviewStatus == ApplyTypes.FinancialReviewStatus.Exempt ?? false;            
            ApplicationResponseViewModel applicationsResponseVm = new ApplicationResponseViewModel();
            applicationsResponseVm.ApplicationResponse = existingApplications;
            applicationsResponseVm.FinancialInfoStage1Expired = financialExpired;
            applicationsResponseVm.FinancialAssessmentUrl = this.Url.Action("StartOrResumeApplication", "Application")
```

src/SFA.DAS.AssessorService.Web/Controllers/Apply/ApplicationController.cs
```diff
@@ -147,7 +158,14 @@ public async Task<IActionResult> StartOrResumeApplication()
         {
             return await StartApplication();
         }
-        
+
+        public IActionResult FinancialExpiredRedir()
```
CodeWolf86, line 162: I don't think this FinancialExpiredRedir is even used?

CodeWolf86 (changes requested):

src/SFA.DAS.AssessorService.Web/Controllers/Apply/ApplicationController.cs
CodeWolf86: Resolving this, after digging I can see it on the form, I was focusing on code behind.

src/SFA.DAS.AssessorService.Web/Controllers/OrganisationStandardsController.cs
CodeWolf86: I've made changes for Uma to Test as @anilkdhiman6  is away and the release was scheduled for Monday. Can talk through when you're back Anil.

src/SFA.DAS.AssessorService.Web/Controllers/Apply/ApplicationController.cs
CodeWolf86: I've made changes for Uma to Test as @anilkdhiman6  is away and the release was scheduled for Monday. Can talk through when you're back Anil.

CodeWolf86 (approved):

CodeWolf86 (approved):

CodeWolf86 (approved):

CodeWolf86 (dismissed):

EwanNoble (dismissed):
Approved for das-sql-permissions as there are no relevant permissions changes

CodeWolf86 (approved):

EwanNoble (approved):

PR #1415, opened 2022-04-14
QF-257 Get Learner endpoint

CodeWolf86:


src/SFA.DAS.AssessorService.Data/LearnerRepository.cs
```diff
@@ -39,6 +39,38 @@ public async Task<Learner> Get(long uln, int stdCode)
...
+                    l.GivenNames,
+                    l.FamilyName,
+                    c.AchievementDate,
+                    c.LatestEPAOutcome as Outcome,
+                    p.Name as ProviderName
+                FROM [Learner] l 
+                    LEFT JOIN [Providers] p on l.UkPrn = p.Ukprn
+                    LEFT JOIN [Certificates] c on l.Uln = c.Uln AND l.StdCode = c.StandardCode 
+                
+                WHERE [ApprenticeshipId] = @apprenticeshipId AND c.Status NOT IN ('Deleted','Draft')",
```
alandburns, line 68: c.Status check needs to be on the JOIN not the WHERE

alandburns (approved):

PR #1420, opened 2022-05-27
SV-1425 Correct validation for opt-in version application

ianatesfa:
Correct validation for opt-in version application

### DIFF
--- a/src/SFA.DAS.AssessorService.Web/Controllers/Apply/StandardController.cs
+++ b/src/SFA.DAS.AssessorService.Web/Controllers/Apply/StandardController.cs
@@ -254,6 +254,9 @@ namespace SFA.DAS.AssessorService.Web.Controllers.Apply
                 return RedirectToAction("Applications", "Application");
 
             var org = await _orgApiClient.GetEpaOrganisation(application.OrganisationId.ToString());
+            if (org == null)
+                return RedirectToAction("Applications", "Application");
+
             IEnumerable<StandardVersion> standards = await _standardVersionApiClient.GetStandardVersionsByIFateReferenceNumber(standardReference);
             StandardVersion stdVersion = standards.FirstOrDefault(x => x.Version.Equals(version, StringComparison.InvariantCultureIgnoreCase));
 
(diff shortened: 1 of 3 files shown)


CodeWolf86 (changes requested):

src/SFA.DAS.AssessorService.Web/Controllers/Apply/StandardController.cs
```diff
@@ -251,39 +251,26 @@ public async Task<IActionResult> OptInPost(Guid id, string standardReference, st
...
-                }
-            }
+            IEnumerable<AppliedStandardVersion> appliedVersions = await _orgApiClient.GetAppliedStandardVersionsForEPAO(org?.OrganisationId, standardReference);
+            AppliedStandardVersion appliedVersion = appliedVersions.FirstOrDefault((x => x.Version.Equals(version, StringComparison.InvariantCultureIgnoreCase)));
+            if (appliedVersion == null || appliedVersion.ApprovedStatus == "Approved" || appliedVersion.ApprovedStatus == "Apply in progress" || appliedVersion.ApprovedStatus == "Feedback Added")
+                return RedirectToAction("Applications", "Application");
 
-            await _orgApiClient.OrganisationStandardVersionOptIn(id, contact.Id, org.OrganisationId, standardReference, version, stdVersion.StandardUId, optInFollowingWithdrawal, $"Opted in by EPAO by {contact.Username}");              
+            DateTime? effectiveTo = appliedVersion.StdVersionEffectiveTo;
+            bool optInFollowingWithdrawal = effectiveTo.HasValue;
```
CodeWolf86, line 271: Can these lines be combined?

bool optInFollowingWithdrawal = appliedVersion.StdVersionEffectiveTo.HasValue 

Don't know if it's worth adding appliedversion.StdVersionEffectiveTo < DateTime.UtcNow.

There are cases where the effective to is set in the future, I'd hope that scenario wouldn't impact here as you shouldn't be able to opt in to a version that you are already approved for, which probably feeds into the GetAppliedVersions logic so ignore this bit. Just typing out loud.

CodeWolf86 (approved):

PR #1440, opened 2022-08-22
Qf 588 move exclusions filter

alandburns:
This change allows the **Active Apprentices by Standard with EPAOs data** to be downloaded as a report without the Exclusions that are applied to the EPAO Opportunity public view of the data. Requires that the StandardSummary data is not filtered for pre-defined Exclusions, but instead the filtering is applied on displaying the standards data in EPAO Opprtunity Finder. New staff report added to Admin service to allow the data to be downloaded without any Exclusions.

### DIFF
--- a/src/SFA.DAS.AssessorService.Database/StoredProcedures/OppFinder_Update_StandardSummary.sql
+++ b/src/SFA.DAS.AssessorService.Database/StoredProcedures/OppFinder_Update_StandardSummary.sql
@@ -18,16 +18,6 @@ DECLARE @Error_Code INT = 0
 BEGIN
 	BEGIN TRANSACTION T1;
 	
-	-- there are some specifically excluded Standards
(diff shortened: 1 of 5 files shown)


src/SFA.DAS.AssessorService.Database/StoredProcedures/OppFinder_Update_StandardSummary.sql
```diff
@@ -168,7 +156,7 @@ BEGIN
 	
 	
 	-- populate the StandardVersionSummary table
-	DELETE FROM StandardVersionSummary WHERE 1=1;
+	TRUNCATE TABLE StandardVersionSummary;
```
CodeWolf86, line 159: Will have to check Truncate Table permissions to make sure this works in opportunity finder.

CodeWolf86 (approved):

PR #1447, opened 2022-11-16
QF-826 Amend PopulateLearner and learner table to handle Transfers

alandburns:
This adds the new transfer logic to Learner table


CodeWolf86:
Superceded by https://github.com/SkillsFundingAgency/das-assessor-service/pull/1449

PR #1455, opened 2023-01-12
QF-930 Run code cleanup. No changes to program logic.

SawNawDfE:


SawNawDfE:
@ChrisJWoodcock It says 969 files changed, but don't worry about the number. Once you use the Hide Whitespace function, the number drops to a paltry 694!

PR #1457, opened 2023-01-19
QF-912 - Further Accessibility Fixes

PeterDavidWilliams:


src/SFA.DAS.AssessorService.Web/Views/Dashboard/Index.cshtml
```diff
@@ -10,6 +10,10 @@
 <main class="govuk-main-wrapper " id="main-content" role="main">
     <div class="govuk-grid-row">
 
+        <h1 class="govuk-heading-xl">
+            Apprenticeship assessment service
```
ChrisJWoodcock, line 14: Did this come from Content or UX design - I'm surprised to see this

ChrisJWoodcock: Need to talk with Victor about a screen design change only for Accessibility which seems wrong and to James to see if the same text as this which is in the header bar could become the h1 which is needed for the accessibility

PeterDavidWilliams: Update with new hidden content per conversation with James King, Jazzy and Victor

ChrisJWoodcock (approved):

ChrisJWoodcock:
This has been merged and combined into a QF-503 branch for easier testing.

PR #1466, opened 2023-02-27
QF-1098 Remove table code

SawNawDfE:


SawNawDfE:
GitHub makes it seem like I added and deleted a bunch of code, but in reality there should be only 8 lines deleted. That is, 4 pairs of 2 table-related HTML tags.

src/SFA.DAS.AssessorService.Web/Views/OrganisationSearch/Results.cshtml
```diff
@@ -57,99 +57,90 @@
...
-                                    }
-                                </td>
-                            </tr>
+                            @("Address not found");
+                        }
+                        else
+                        {
+                            var addressArray = new[] { @result.Address?.Address1, @result.Address?.City, @result.Address.Postcode };
+                            @string.Join(", ", addressArray.Where(s => !string.IsNullOrEmpty(s)));
+                        }
```
SawNawDfE, line 84: I don't see how this would have ever worked. The return value of `string.Join()` is not being used!

I didn't add this code, despite GitHub's green highlighting.

ChrisJWoodcock: Doesn't it just drop the value onto the screen as output?

SawNawDfE: Ah yes, this is razor

ChrisJWoodcock (approved):

PR #1468, opened 2023-03-15
QF-625 Upgrade to .NET 6 and upgrade some packages

SawNawDfE:
This is a new PR for the upgraded YAML pipelines.


ChrisJWoodcock (approved):
This is a replacement PR to attempt to workaround the issue with the name of the pipeline having changed from "das-assessor-service YAML Migration" to "das-assessor-service"

UPDATE: The problem with the original PR was traced to the 'Branch Protection Rules' these have been corrected to remove the old pipeline name.

Mostly the checks are passing the Sonar Cloud issues are being addressed on separate branches.

PR #1470, opened 2023-03-17
QF-977 Split assessor web config

ChrisJWoodcock:


src/SFA.DAS.AssessorService.Web.UnitTests/MockedObjects/MockedCertificateApiClient.cs
```diff
@@ -1,28 +1,23 @@
...
-            var webConfigMock = new Mock<IWebConfiguration>();
-            var hostMock = new Mock<IHostEnvironment>();
-            hostMock
-                .Setup(m => m.EnvironmentName)
-                .Returns(Environments.Development);
-            var tokenServiceMock = new TokenService(webConfigMock.Object, hostMock.Object, false);
+            var tokenServiceMock = new Mock<ITokenService>();
+            tokenServiceMock
+                .Setup(m => m.GetToken())
+                .Returns(string.Empty);
```
SawNawDfE, line 20: Wouldn't it be better to mock the `IHostEnvironment`'s property like the old code was doing? If for whatever reason we want the development environment to use a token in the future, that would cause this new test to fail

src/SFA.DAS.AssessorService.Web/Startup.cs
```diff
@@ -143,40 +162,43 @@ private IServiceProvider ConfigureIoc(IServiceCollection services)
...
+                config.For<IApprovalsLearnerApiClient>().Use<ApprovalsLearnerApiClient>().Ctor<ITokenService>().Is(c => c.GetInstance<ITokenService>("AssessorTokenService"));
+                config.For<ILocationsApiClient>().Use<LocationsApiClient>().Ctor<ITokenService>().Is(c => c.GetInstance<ITokenService>("AssessorTokenService"));
+                config.For<IStandardVersionClient>().Use<StandardVersionClient>().Ctor<ITokenService>().Is(c => c.GetInstance<ITokenService>("AssessorTokenService"));
+
+                config.For<IQnaApiClient>().Use<QnaApiClient>().Ctor<ITokenService>().Is(c => c.GetInstance<ITokenService>("QnATokenService"));
+                config.For<IRoatpApiClient>().Use<RoatpApiClient>().Ctor<ITokenService>().Is(c => c.GetInstance<ITokenService>("RoatpTokenService"));
+                config.For<IAzureApiClient>().Use<AzureApiClient>().Ctor<string>().Is(Configuration.AzureApiAuthentication.ApiBaseAddress);
 
+                config.For<IApiValidationService>().Use<ApiValidationService>().Ctor<ITokenService>().Is(c => c.GetInstance<ITokenService>("AssessorTokenService"));
```
SawNawDfE, line 201: Since there are a dozen or so references to "AssessorTokenService", they can be replaced with a reference to a `private const string` instead of being hardcoded

src/SFA.DAS.AssessorService.Settings/ConfigurationService.cs
```diff
@@ -8,7 +8,26 @@ namespace SFA.DAS.AssessorService.Settings
...
+            return config;
+        }
+
+        public static async Task<IWebConfiguration> GetConfigWeb(string environment, string storageConnectionString, string version, string serviceName)
+        {
+            var config = await GetConfig <WebConfiguration>(environment, storageConnectionString, version, serviceName);
+            return config;
+        }
+
+        private static async Task<T> GetConfig<T>(string environment, string storageConnectionString, string version, string serviceName)
```
SawNawDfE, line 30: We should have some tests for this new code that returns differernt types of configs

SawNawDfE (changes requested):

src/SFA.DAS.AssessorService.Web/Startup.cs
ChrisJWoodcock: Done that and the same in the API

SawNawDfE (dismissed):

PeterDavidWilliams (approved):

debrr07 (approved):

PeterDavidWilliams (approved):

PR #1472, opened 2023-03-24
QF-1202 Prevent Cross Site Scripting in Change Address fields

SawNawDfE:
- Prevented HTML tags from passing validation in address fields
- Removed the System.ServiceModel.Federation floating package version that was added by Visual Studio while upgrading the Connected Services during the .NET 6 package. This was causing a version mismatch.
- Updated and consolidated packages related to unit testing
- Updated and consolidated NLog packages
- Updated and consolidated Swashbuckle packages


SawNawDfE:
I will put up a new PR with the main code changes done in JavaScript - closing this.

PR #1473, opened 2023-03-24
QF-1181 Lower log level from error to info on ULN - Standard mismatches

PeterDavidWilliams:
Also added to existing unit test to confirm correct log level and message is logged


src/SFA.DAS.AssessorService.Application.Api/SFA.DAS.AssessorService.Application.Api.csproj
```diff
@@ -29,9 +29,9 @@
     <PackageReference Include="Swashbuckle.AspNetCore.Swagger" Version="6.5.0" />
     <PackageReference Include="System.ServiceModel.Duplex" Version="4.10.0" />
     <PackageReference Include="System.ServiceModel.Federation" Version="4.10.*" />
```
ChrisJWoodcock, line 31: Remove package System.ServiceModel.Federation and restore others to 4.10.0

ChrisJWoodcock (changes requested):

ChrisJWoodcock (approved):

PR #1479, opened 2023-04-04
QF-1203 Resolve sonarcloud security vulnerability

SarahDuncan:


ChrisJWoodcock (changes requested):

src/SFA.DAS.AssessorService.Data/Staff/StaffReportRepository.cs
```diff
@@ -42,10 +41,23 @@ public Task<IEnumerable<StaffReport>> GetReportList()
...
+                {
+                    return null;
+                }
+
+                return await GetReport(reportId.Id);
+            }
+            catch
+            {
+                return null;
+            }
```
SawNawDfE, line 60: A rule of thumb is to avoid catching exceptions if you don't have a good way of handling them. Here, not only is the catch block not handling the error, it's suppressing it and preventing its caller from catching SqlExceptions (on line 96 of StaffReportsController).

I think we should allow it to visibly blow up without catching the exception.

SawNawDfE (changes requested):

src/SFA.DAS.AssessorService.Application.Api/Controllers/Staff/StaffReportsController.cs
```diff
@@ -115,7 +115,7 @@ public async Task<IActionResult> GetReport(Guid reportId)
             }
             catch (SqlException sqlEx)
             {
-                _logger.LogInformation($"Could not generate report due to : {sqlEx.Message}");
+                _logger.LogError($"Could not generate data from report Id [{reportId}] due to : {sqlEx.Message}");
                 return NoContent();
```
SawNawDfE, line 119: Here we're logging an error, but returning a success status. That doesn't seem consistent
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204

SawNawDfE (changes requested):

src/SFA.DAS.AssessorService.Application.Api/Controllers/Staff/StaffReportsController.cs
SarahDuncan: @ChrisJWoodcock

SarahDuncan:
Something went wrong with the pull request ( it has many random commits in it from master), so a branch and pull request has been made.

src/SFA.DAS.AssessorService.Application.Api/Controllers/Staff/StaffReportsController.cs
SawNawDfE: Resolving this due PR closure

PR #1484, opened 2023-04-25
QF-1203 Resolve security issue

SarahDuncan:
Edit code to be more robust against malicious inputs when running a stored proc against the database.

### DIFF
--- a/src/SFA.DAS.AssessorService.Application.Api/Controllers/Staff/StaffReportsController.cs
+++ b/src/SFA.DAS.AssessorService.Application.Api/Controllers/Staff/StaffReportsController.cs
@@ -1,7 +1,6 @@
 ﻿using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Logging;
-using Microsoft.IdentityModel.Tokens;
 using SFA.DAS.AssessorService.Api.Types.Models.AO;
 using SFA.DAS.AssessorService.Application.Api.Middleware;
 using SFA.DAS.AssessorService.Application.Api.Properties.Attributes;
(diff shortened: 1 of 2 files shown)


src/SFA.DAS.AssessorService.Application.Api/Controllers/Staff/StaffReportsController.cs
```diff
@@ -115,7 +115,7 @@ public async Task<IActionResult> GetReport(Guid reportId)
             }
             catch (SqlException sqlEx)
             {
-                _logger.LogInformation($"Could not generate report due to : {sqlEx.Message}");
+                _logger.LogError($"Could not generate data from report Id [{reportId}] due to : {sqlEx.Message}");
                 return NoContent();
```
SawNawDfE, line 119: Here we're logging an error, but returning a success status. That doesn't seem consistent
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204

SarahDuncan: @ChrisJWoodcock  please see above

src/SFA.DAS.AssessorService.Data/Staff/StaffReportRepository.cs
```diff
@@ -45,7 +44,14 @@ public Task<IEnumerable<StaffReport>> GetReportList()
 
         public async Task<IEnumerable<IDictionary<string, object>>> GetDataFromStoredProcedure(string storedProcedure)
         {
-            return (await _connection.QueryAsync(storedProcedure, commandType: CommandType.StoredProcedure)).OfType<IDictionary<string, object>>().ToList();
+            var reportId = _assessorDbContext.StaffReports.FirstOrDefault(rep => rep.StoredProcedure == storedProcedure.Trim());
```
ChrisJWoodcock, line 47: Should rename to report

ChrisJWoodcock (changes requested):

src/SFA.DAS.AssessorService.Data/Staff/StaffReportRepository.cs
SarahDuncan: sorted

ChrisJWoodcock (approved):

PR #1487, opened 2023-05-09
QF-1348 Using background task queue for long running operation

ChrisJWoodcock:


PeterDavidWilliams (approved):

src/SFA.DAS.AssessorService.Application.Api/Controllers/BaseController.cs
```diff
@@ -0,0 +1,40 @@
+﻿using MediatR;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Logging;
+using SFA.DAS.AssessorService.Api.Types.Models;
+using SFA.DAS.AssessorService.Application.Api.TaskQueue;
+using System;
+
+namespace SFA.DAS.AssessorService.Application.Api.Controllers
+{
+    public class BaseController : Controller
```
Jordan-fujitsu, line 10: @ChrisJWoodcock `BaseController`  name seems a bit of a vague name, do you think there's more specific name which could be used like e.g `BackgroundQueueBaseController ` or something similar

ChrisJWoodcock: I had in mind it really being the start of a BaseController for all controllers

Jordan-fujitsu (approved):

Jordan-fujitsu (approved):

PR #1503, opened 2023-06-15
QF-1336 database schema only changes for EPAO build

ChrisJWoodcock:


ChrisJWoodcock (commented):
This PR contains only the schema changes so that we can swap branches more easily in the TEST environments by allowing these tables into master

ChrisJWoodcock:
Abandoning this PR as there are schema changes on the other tickets which cannot be merged to master e.g. the column type in OrganisationStandardVersions table for the comments, so there is no benefit to these changes going into master, the environment problems caused by the schema changes will have to be dealt with manually.

PR #1505, opened 2023-06-15
QF-1460 - Turn off new organisation apply

PeterDavidWilliams:
… database for organisations


ChrisJWoodcock (changes requested):
There is also OrganisationSearchByUkprn - which needs the same commenting out

ChrisJWoodcock (approved):

PR #1510, opened 2023-06-21
QF-602 Change given names and family name on UI

SarahDuncan:


src/SFA.DAS.AssessorService.Application.Api.UnitTests/Services/CertificateNameCapitalisationServiceTests.cs
```diff
@@ -0,0 +1,163 @@
+﻿using Moq;
+using NUnit.Framework;
+using SFA.DAS.AssessorService.Application.Api.Services;
+
+namespace SFA.DAS.AssessorService.Application.Api.UnitTests.Services
+{
+    [TestFixture]
+    public class CertificateNameCapitalisationServiceTests
+    {
+        private Mock<CertificateNameCapitalisationService> _certificateNameCapitalisationService;
```
ChrisJWoodcock, line 10: Since the service is being tested here you would instantiate the actual class not a Mock<T>

src/SFA.DAS.AssessorService.Application.Api.UnitTests/Services/CertificateNameCapitalisationServiceTests.cs
```diff
@@ -0,0 +1,163 @@
...
+
+        [SetUp]
+        public void Setup()
+        {
+            _certificateNameCapitalisationService = new Mock<CertificateNameCapitalisationService>();
+        }
+
+        [TestCase(" Test")]
```
ChrisJWoodcock, line 18: [TestCase(" Test", "Test")] it is better to give the specific expected results rather than rely on a method.

src/SFA.DAS.AssessorService.Application.Api.UnitTests/Services/CertificateNameCapitalisationServiceTests.cs
```diff
@@ -0,0 +1,163 @@
...
+            var actualName = _certificateNameCapitalisationService.Object.ProperCase(inputName, true);
+
+            Assert.AreEqual(expectedName, actualName);
+        }
+
+        [TestCase("smith")]
+        [TestCase("SMITH")]
+        [TestCase("pienaar")]
+        [TestCase("PIENAAR")]
+        public void AndNameIsInAllLowercaseOrUppercase_ThenTheNameIsCapitalisedCorrectly(string name)
```
ChrisJWoodcock, line 60: Specify the expected result in the TestCase

src/SFA.DAS.AssessorService.Application.Api/Services/CertificateNameCapitalisationService.cs
```diff
@@ -0,0 +1,199 @@
...
+            }
+
+            return namePart;
+        }
+
+        /// <summary>
+        /// Cleans all magic characters
+        /// </summary>
+        /// <returns></returns>
+        public void CleanseStringForSpecialCharacters(ref string inputString)
```
ChrisJWoodcock, line 119: Check if these methods are used and make private if they are not and remove from interface

src/SFA.DAS.AssessorService.Application.UnitTests/Handlers/Certificates/StartCertificateHandlerTests/StartCertificateHandlerTests.cs
```diff
@@ -55,6 +55,8 @@ public void SetUp()
             // Arrange
             request.StandardUId = null;
             learnerRecord.FundingModel = 81;
+            learnerRecord.GivenNames = CapitaliseFirstLetterOfName(learnerRecord.GivenNames);
```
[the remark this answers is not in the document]
ChrisJWoodcock: Now that you have _mockCertificateNameCapitalisationService you can mock the ProperCase to just return it's input which would be more correct since you do not care what it does either way

src/SFA.DAS.AssessorService.Web.UnitTests/CertificateTests/CertificateFamilyNameControllerTests.cs
```diff
@@ -0,0 +1,110 @@
...
+
+        [Test, MoqAutoData]
+        public async Task WhenRequestingFamilyNameView_ThenLoadsFamilyNameViewSuccessfully(CertificateSession session)
+        {
+            var sessionString = JsonConvert.SerializeObject(session);
+            _mockSessionService.Setup(s => s.Get(nameof(CertificateSession))).Returns(sessionString);
+
+            var result = await _controller.FamilyName() as ViewResult;
+
+            result.ViewName.Should().Be("~/Views/Certificate/FamilyName.cshtml");
```
ChrisJWoodcock, line 72: using (new AssertionScope()) {} can be used to group multiple checks in FluentAssertions similar to Assert.Multiple

src/SFA.DAS.AssessorService.Web/Controllers/CertificateBaseController.cs
```diff
@@ -55,9 +55,17 @@ protected async Task<IActionResult> LoadViewModel<T>(string view) where T : Cert
 
             Logger.LogInformation($"Got View Model of type {typeof(T).Name} requested by {username}");
 
+            if (!returnView && viewModel.GetType() == typeof(CertificateNamesViewModel)) { SetNamePropertyAndReturnView(viewModel as CertificateNamesViewModel); }
```
ChrisJWoodcock, line 58: This is a red flag - a base class needing to know the type of one of its derived classes

src/SFA.DAS.AssessorService.Web/Validators/CertificateNameViewModelValidator.cs
```diff
@@ -0,0 +1,21 @@
+﻿using FluentValidation;
+using SFA.DAS.AssessorService.Web.ViewModels.Certificate;
+
+namespace SFA.DAS.AssessorService.Web.Validators
```
ChrisJWoodcock, line 4: There is a problem with this, the previous name is being compared to the new name to see if only the case has differed - but as the previous name is being stored in a hidden field it is possible to construct a POST such that you change both the previous and the old names and allow yourself to change the name to anything - the previous name should not be stored in the view instead the certificate id should be stored and then in the validation rule lookup the certificate current name from the database and do the comparison to see if only the case has changed between that and the new name

ChrisJWoodcock (changes requested):

src/SFA.DAS.AssessorService.Application.UnitTests/Handlers/Certificates/StartCertificateHandlerTests/StartCertificateHandlerTests.cs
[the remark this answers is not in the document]
SarahDuncan: I have not done this because the algorithm should not run _at all_ in these tests. Setting it up the return the input value conflicts with the fact that it shouldn't run at all.

ChrisJWoodcock (approved):

PR #1526, opened 2023-07-10
QF-1540 Standard apply with eqap name simplified questions

ChrisJWoodcock:


src/SFA.DAS.AssessorService.Web/Controllers/Apply/StandardController.cs
```diff
@@ -241,182 +243,200 @@ public async Task<IActionResult> StandardDetails(string referenceNumber)
...
         {
-            var contact = await GetUserContact();
-            var org = await _orgApiClient.GetOrganisationByUserId(contact.Id);
+            if (id == Guid.Empty)
+                throw new ArgumentException($"Value of {nameof(id)} cannot be empty");
 
-            var existingApplications = (await _applicationApiClient.GetStandardApplications(contact.Id))?
-                .Where(p => p.ApplicationStatus != ApplicationStatus.Declined);
+            if (string.IsNullOrEmpty(search))
+                throw new ArgumentException($"Value of {nameof(search)} cannot be null or empty");
```
SawNawDfE, line 291: [SonarCloud makes a good point](https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&pullRequest=1526&id=SkillsFundingAgency_das-assessor-service&open=AYlAixI_iks5oidIUBLY&tab=code) about the newly added parameter checks in these async methods. 

The only things I would add is that exceptions being far-removed in time from the actual error would make the logs more confusing, and to suggest using guard clauses in place of this

src/SFA.DAS.AssessorService.Web/Controllers/Apply/StandardController.cs
```diff
@@ -241,182 +243,200 @@ public async Task<IActionResult> StandardDetails(string referenceNumber)
...
+            if (standardVersions.Any(x => x.ApprovedStatus == ApprovedStatus.Approved))
             {
-                selectedStandard = standardVersions.LastOrDefault();
-                versions = model.SelectedVersions ?? new List<string> { selectedStandard.Version };
-                if (model.SelectedVersions != null)
-                    applicationStatus = await ApplicationStandardStatus(application, standardReference, model.SelectedVersions);
+                return RedirectToRoute(StandardDetailsRouteGet, new { referenceNumber });
             }
-            else
+            else if(latestStandard.EqaProviderName == "Ofqual")
```
SawNawDfE, line 373: We should account for latestStandard being null here

src/SFA.DAS.AssessorService.Web/Validators/Apply/ApplyStandardConfirmViewModelValidator.cs
```diff
@@ -0,0 +1,19 @@
...
+using SFA.DAS.AssessorService.Web.ViewModels.Standard;
+
+namespace SFA.DAS.AssessorService.Web.Validators.Standard
+{
+    public class ApplyStandardConfirmViewModelValidator : AbstractValidator<ApplyStandardConfirmViewModel>
+    {
+        public ApplyStandardConfirmViewModelValidator()
+        {
+            RuleFor(vm => vm.IsConfirmed)
+                .NotEmpty()
```
SawNawDfE, line 11: Not sure how this could be empty -- isn't this a non-nullable boolean?

src/SFA.DAS.AssessorService.Web/Validators/Apply/ApplyStandardConfirmViewModelValidator.cs
```diff
@@ -0,0 +1,19 @@
...
+                .WithMessage("Confirm you have read the assessment plan");
+
+            RuleFor(vm => vm.SelectedVersions)
+                .NotEmpty()
```
SawNawDfE, line 15: Can the collection be null?

src/SFA.DAS.AssessorService.Web/Validators/Standard/AddStandardSearchViewModelValidator.cs
```diff
@@ -3,12 +3,12 @@
...
-    public class AddStandardViewModelValidator : AbstractValidator<AddStandardSearchViewModel>
+    public class AddStandardSearchViewModelValidator : AbstractValidator<AddStandardSearchViewModel>
     {
-        public AddStandardViewModelValidator()
+        public AddStandardSearchViewModelValidator()
         {
             const string searchMessage = "Enter 3 or more characters to search";
-            RuleFor(vm => vm.StandardToFind)
+            RuleFor(vm => vm.Search)
                 .NotEmpty().WithMessage(searchMessage)
```
SawNawDfE, line 12: Think this check should happen after a `.Trim()` to account for whitespace

SawNawDfE (changes requested):
Left a few comments

src/SFA.DAS.AssessorService.Web/Validators/Apply/ApplyStandardConfirmViewModelValidator.cs
ChrisJWoodcock: This does work as it compares to the default value of the type e.g. false - however better to make that explicit so done that

src/SFA.DAS.AssessorService.Web/Controllers/Apply/StandardController.cs
ChrisJWoodcock: Split methods are per Sonar Cloud suggestions

src/SFA.DAS.AssessorService.Web/Controllers/Apply/StandardController.cs
ChrisJWoodcock: Added handling for various cases

src/SFA.DAS.AssessorService.Web/Validators/Standard/AddStandardSearchViewModelValidator.cs
ChrisJWoodcock: Done that in both cases

src/SFA.DAS.AssessorService.Web/Validators/Apply/ApplyStandardConfirmViewModelValidator.cs
SawNawDfE: Hmm, that's some unintuitive behaviour on the library's part considering that empty and false are two very different things

PR #1534, opened 2023-07-13
QF-1473 - Removed assessor link from find an opportunity

PeterDavidWilliams:


SarahDuncan (approved):

src/SFA.DAS.AssessorService.Web/Views/OppFinder/Index.cshtml
```diff
@@ -105,9 +105,6 @@
             <p class="govuk-body govuk-!-margin-bottom-1">
                 <a class="govuk-link" href="https://www.gov.uk/guidance/how-apprentices-in-england-are-assessed" target="_blank" rel="noopener noreferrer">How apprentices in England are assessed</a>
             </p>
-            <p class="govuk-body">
```
ChrisJWoodcock, line 108: You have missed AC2 and AC3?

ChrisJWoodcock (changes requested):

src/SFA.DAS.AssessorService.Web/Views/OppFinder/Index.cshtml
PeterDavidWilliams: The three ACs share the same base page, so removing it from the Index.cshtml page satisfies all three ACs. I've checked it locally to confirm this is the case, happy to show you

PeterDavidWilliams: Turns out I did miss one not covered by the ACs. Corrected

PR #1536, opened 2023-07-14
QF-1541 Disable the Add Standard button

SawNawDfE:


src/SFA.DAS.AssessorService.Web/Views/OrganisationStandards/Index.cshtml
```diff
@@ -113,7 +113,7 @@
         </div>
     }
 
-    <a asp-route="@StandardController.AddStandardSearchRouteGet" class="govuk-button">Add standard</a>
+    <!-- <a asp-route="@StandardController.AddStandardSearchRouteGet" class="govuk-button">Add standard</a> -->
```
SawNawDfE, line 116: I haven't deleted this entirely to make it more convenient for ourselves in case we want to re-enable it tomorrow

ChrisJWoodcock (approved):

PR #1539, opened 2023-07-20
QF-1556 Update banners for end july

ChrisJWoodcock:


src/SFA.DAS.AssessorService.Web/Views/Banners/_NoAssessBannerPartial.cshtml
```diff
@@ -7,7 +7,7 @@
         </h2>
     </div>
     <div class="govuk-notification-banner__content">
-        <p class="govuk-notification-banner__heading">From 14 July to 31 July 2023 you will not be able to apply to assess new standards or versions.</p>
-        <p>New end-point assessment organisations will be able to use this service from 1 August 2023.</p>
+        <p class="govuk-notification-banner__heading">The Register of end-point assessment organisations (EPAO) is now part of the Apprenticeship provider and assessment register (APAR).</p>
+        <p>If your organisation was previously on the Register of EPAOs, your details and approved standards have been transferred to APAR. You can sign in to the apprenticeship assessment service with your existing account.</p>
```
SawNawDfE, line 11: Since the name of this partial view no longer makes sense, we should rename it

SawNawDfE (changes requested):

src/SFA.DAS.AssessorService.Web/Views/Banners/_NoAssessBannerPartial.cshtml
ChrisJWoodcock: Done

SawNawDfE (approved):

PR #1553, opened 2023-08-02
QF-1566 Test OFQUAL-to-EpAO Merge

ChrisJWoodcock:


src/SFA.DAS.AssessorService.Data.IntegrationTests/Models/StagingOfqualStandardModel.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace SFA.DAS.AssessorService.Data.IntegrationTests.Models
+{
+    public class StagingOfqualStandardModel : TestModel
+    {
+        public string RecognitionNumber { get; set; }
+
+        public DateTime OperationalStartDate { get; set; }
+
+        public DateTime? OperationalEndDate { get; set; } // Nullable DateTime
```
SawNawDfE, line 11: This comment can go

SawNawDfE (dismissed):

src/SFA.DAS.AssessorService.Database/StoredProcedures/Load_OFQUAL_Standards.sql
```diff
@@ -50,7 +50,26 @@ BEGIN
...
+                  ,[Website]
+                  ,[HeadOfficeAddressLine1]
+                  ,[HeadOfficeAddressLine2]
+                  ,[HeadOfficeAddressTown]
+                  ,[HeadOfficeAddressCounty]
+                  ,[HeadOfficeAddressPostcode]
+                  ,[HeadOfficeAddressCountry]
+                  ,[HeadOfficeAddressTelephone]
+                  ,[OfqualStatus]
+                  ,ISNULL([OfqualRecognisedFrom], '1900-01-01 00:00:00.000') OfqualRecognisedFrom
```
SawNawDfE, line 70: Will it be an issue that the time has 7 digits after the dot instead of 3? As in `2013-09-17 00:00:00.0000000`

SawNawDfE (approved):

src/SFA.DAS.AssessorService.Database/StoredProcedures/Load_OFQUAL_Standards.sql
ChrisJWoodcock: I don't think so - it seems to work anyway

ChrisJWoodcock (commented):
This has been approved by another developer but they are non in the SQL Permissions role which is preventing this being merged.

I have asked another Lead Developer to approve on my behalf

PR #1558, opened 2023-08-23
QF-1610 Store CoronationEmblem field 

SawNawDfE:


ChrisJWoodcock (approved):

SawNawDfE:
@ChrisJWoodcock I pushed some changes to this branch that shouldn't have been pushed. I've since reverted them -- please could you re-approve?

src/SFA.DAS.AssessorService.Database/Tables/Standards.sql
```diff
@@ -6,6 +6,7 @@
 	[Title] VARCHAR(1000) NOT NULL,
 	[Version] VARCHAR(20) NULL,
 	[Level] INT NOT NULL,
+	[CoronationEmblem] BIT NOT NULL DEFAULT 0,
```
ChrisJWoodcock, line 9: This column should also move to the end of the list to avoid the table rebuilding in the DacPac deployment

ChrisJWoodcock (changes requested):

src/SFA.DAS.AssessorService.Database/Tables/Standards.sql
SawNawDfE: Done

ChrisJWoodcock (dismissed):

ChrisJWoodcock (approved):

PR #1559, opened 2023-08-25
QF-1611 Add coronation emblem to certificates

SawNawDfE:


src/SFA.DAS.AssessorService.Application/Handlers/Certificates/UpdateCertificateHandler.cs
```diff
@@ -54,7 +56,10 @@ public async Task<Certificate> Handle(UpdateCertificateRequest request, Cancella
                 _logger.LogInformation($"Certificate with ID: {request.Certificate.Id} Updated with reference of {request.Certificate.CertificateReference}");
             }
 
-            updatedData = HandleSendToUpdate(currentCertificate, currentData, updatedData);
+            updatedData = CertificateDataSendToUpdater.HandleSendToUpdate(currentCertificate, currentData, updatedData);
+
+            updatedData.CoronationEmblem = await _standardRepository.GetCoronationEmblemForStandardReferenceAndVersion(updatedData.StandardReference, updatedData.Version);
+            updatedData.StandardName = await _standardRepository.GetTitleForStandardReferenceAndVersion(updatedData.StandardReference, updatedData.Version);
```
SawNawDfE, line 62: This gets the CoronationEmblem and StandardName from the database every time the certificate is updated. This fulfills ACs 3.2 and 3.3, and I find it neater to have the Assessor decide this instead of having the Admin Service supply the correct values.

ChrisJWoodcock: There should be an equivalent class to the CertificateDataSentToUpdater i.e. CertficateDataVersionUpdater which checks if the version has changed and then does these updates as that is the pattern to follow and these update should only be done when the version has changed which the AC's do say

ChrisJWoodcock (changes requested):

src/SFA.DAS.AssessorService.Application.Api.UnitTests/Controllers/Standards/StandardVersionControllerTests.cs
```diff
@@ -45,7 +44,7 @@ public async Task WhenRequestingGetAllStandardVersions_ThenListOfStandardVersion
 
             var model = controllerResult.Value as IEnumerable<StandardVersion>;
 
-            var expectedResponse = standards.Select(ConvertFromStandard);
+            var expectedResponse = standards.Select(s => (StandardVersion)s);
```
SawNawDfE, line 47: I've fixed this and other similar failing tests by using the existing implicit operator in the `StandardVersion` class instead of duplicating the mapping logic inside the test itself.

src/SFA.DAS.AssessorService.Application/Handlers/Certificates/UpdateCertificateHandler.cs
SawNawDfE: Done. I've also fixed the existing tests in several places

ChrisJWoodcock (changes requested):

ChrisJWoodcock (approved):

src/SFA.DAS.AssessorService.Data/StandardRepository.cs
```diff
@@ -680,7 +680,21 @@ public async Task<bool> GetCoronationEmblemForStandardReferenceAndVersion(string
...
+            var sql = @"SELECT [CoronationEmblem]
+                       FROM [Standards] 
+                       WHERE [StandardUId] = @standardUId";
+
+            var results = await _unitOfWork.Connection.QueryAsync<bool>(
+                sql,
+                param: new { standardUId },
+                transaction: _unitOfWork.Transaction);
+
+            return results.Single();
```
SawNawDfE, line 697: I've added this and used it in the StartCertificateHandler. At the start of the Record a Grade journey, new certificates were being generated using data from the wrong standard version, because certificate data for new certificates is retrieved from the latest standard version regardless of the actual version being graded.  

With this commit, the value of CoronationEmblem will be retrieved from the correct standard version. I raised a [bug ](https://skillsfundingagency.atlassian.net/browse/QF-1633)to address the other fields.
@ChrisJWoodcock

src/SFA.DAS.AssessorService.Data/StandardRepository.cs
```diff
@@ -667,5 +668,47 @@ private DataTable ConstructStandardOptionsDataTable(IEnumerable<StandardOption>
...
```
SawNawDfE, line 697: I've added this and used it in the StartCertificateHandler. At the start of the Record a Grade journey, new certificates were being generated using data from the wrong standard version, because certificate data for new certificates is retrieved from the latest standard version regardless of the actual version being graded.

With this commit, the value of CoronationEmblem will be retrieved from the correct standard version. I raised a [bug ](https://skillsfundingagency.atlassian.net/browse/QF-1633)to address the other fields.
@ChrisJWoodcock

ChrisJWoodcock (approved):

PR #1580, opened 2023-11-23
QF-1731-Ofqual-contact-details-not-pulling-through-into-opportunity-finder-BUG

Jordan-fujitsu:
Adding contact name to sproc and adding null/empty string check

### DIFF
--- a/src/SFA.DAS.AssessorService.Application.UnitTests/Handlers/OppFinder/WhenHandlingGetOppFinderApprovedStandardDetailsRequest.cs
+++ b/src/SFA.DAS.AssessorService.Application.UnitTests/Handlers/OppFinder/WhenHandlingGetOppFinderApprovedStandardDetailsRequest.cs
@@ -25,11 +25,6 @@ namespace SFA.DAS.AssessorService.Application.UnitTests.Handlers.OppFinder
         private GetOppFinderApprovedStandardDetailsHandler _sut;
 
 
-        //[OneTimeSetUp]
-        //public void OneTimeSetup() 
-        //{
-            
-        //}
 
         [SetUp]
         public async Task SetupAsync()
@@ -104,21 +99,6 @@ namespace SFA.DAS.AssessorService.Application.UnitTests.Handlers.OppFinder
             }
         }
 
-        [Test]
-        public async Task ThenReturnsEmails()
-        {
-            //Arrange
-            var request = new GetOppFinderApprovedStandardDetailsRequest()
-            {
-                StandardReference = "ST0001"
-            };
-
-            //Act
-            var result = await _sut.Handle(request, new CancellationToken());
-
-            //Assert
-            result.EqaProvider.Should().Be("EMAIL");
-        }
 
         [Test]
         public async Task ThenReturnsRegionDetails()
--- a/src/SFA.DAS.AssessorService.Application.UnitTests/Handlers/OppFinder/WhenHandlingGetOppFinderApprovedStandardDetailsRequest.cs
+++ b/src/SFA.DAS.AssessorService.Application.UnitTests/Handlers/OppFinder/WhenHandlingGetOppFinderApprovedStandardDetailsRequest.cs
@@ -147,6 +147,7 @@ namespace SFA.DAS.AssessorService.Application.UnitTests.Handlers.OppFinder
         [TestCase("eqaProviderContactName@email.com", "", "eqaProviderContactName@email.com", "OFQUAL", "eqaProviderContactName@email.com")]
         [TestCase("", "", "", "OFQUAL", "OFQUAL")]
         [TestCase("eqaProviderContactName", "", "eqaProviderContactName", "OFQUAL", "OFQUAL")]
+        [TestCase("eqaProviderContactEmail", "eqaProviderContactEmail", "", "OFQUAL", "OFQUAL")]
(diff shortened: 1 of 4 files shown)


src/SFA.DAS.AssessorService.Application.UnitTests/Handlers/OppFinder/WhenHandlingGetOppFinderApprovedStandardDetailsRequest.cs
```diff
@@ -117,33 +134,60 @@ public async Task ThenReturnsRegionDetails()
...
                 StandardReference = "ST0001"
             };
 
+            var validationRequest = new ValidationRequest()
+            {
+                Type = "email",
+                Value = validationRequestEmail
+            };
+
+            ValidationHandler validationHandler = new ValidationHandler(new ValidationService());
```
ChrisJWoodcock, line 165: You should be mocking the validation handler instead of using the real one so mock for "" returns true, "valid@email.com" returns true and "invalid" returns false, then use those values for the inputs, otherwise you are dependent on the implementation of the validation handler in this test

src/SFA.DAS.AssessorService.Application.UnitTests/Handlers/OppFinder/WhenHandlingGetOppFinderApprovedStandardDetailsRequest.cs
```diff
@@ -117,33 +134,60 @@ public async Task ThenReturnsRegionDetails()
...
             var result = await _sut.Handle(request, new CancellationToken());
 
             //Assert
-            result.VersionResults.Count.Should().Be(2);
-            using (new AssertionScope())
-            {
-                result.VersionResults[0].Version.Should().Be("1.0");
-                result.VersionResults[1].Version.Should().Be("1.1");
-            }
+            result.Should().NotBeNull();
```
ChrisJWoodcock, line 189: When doing more than 1 fluid assertion using an AssertionScope is best practice or remove the NotBeNull as the test will fail if it by throwing an exception on the second assert

src/SFA.DAS.AssessorService.Application/Handlers/OppFinder/GetOppFinderApprovedStandardDetailsHandler.cs
```diff
@@ -35,11 +35,11 @@ public async Task<GetOppFinderApprovedStandardDetailsResponse> Handle(GetOppFind
 
             string eqaProvider;
```
ChrisJWoodcock, line 37: Better to use var eqaProviderContactEmail = result.OverviewResult?.EqaProviderContactEmail; instead of duplicating the path in one statement

src/SFA.DAS.AssessorService.Application.UnitTests/Handlers/OppFinder/WhenHandlingGetOppFinderApprovedStandardDetailsRequest.cs
```diff
@@ -22,23 +24,31 @@ public class WhenHandlingGetOppFinderApprovedStandardDetailsRequest
...
             _mockLogger = new Mock<ILogger<GetOppFinderApprovedStandardDetailsHandler>>();
+            var validationRequest = new ValidationRequest()
+            {
+                Type = "email",
+                Value = "EMAIL",
+            };
 
-            _mockMediator.Setup(s => s.Send(It.Is<ValidationRequest>(x => x.Type == "email" && x.Value == "EMAIL"), It.IsAny<CancellationToken>()))
-                        .ReturnsAsync(true);
+            ValidationHandler validationHandler = new ValidationHandler(new ValidationService());
```
ChrisJWoodcock, line 41: See comment below first - but this would just be a complicted It.IsAny return false from a mock - which is strange as the original returned true - is it really being used?

ChrisJWoodcock (changes requested):

src/SFA.DAS.AssessorService.Application.UnitTests/Handlers/OppFinder/WhenHandlingGetOppFinderApprovedStandardDetailsRequest.cs
Jordan-fujitsu: Updated

src/SFA.DAS.AssessorService.Application.UnitTests/Handlers/OppFinder/WhenHandlingGetOppFinderApprovedStandardDetailsRequest.cs
Jordan-fujitsu: Removed validation handler and instead mocked with valid values being "", null, "valid@email.com" and invalid being "invalid"

src/SFA.DAS.AssessorService.Application.UnitTests/Handlers/OppFinder/WhenHandlingGetOppFinderApprovedStandardDetailsRequest.cs
Jordan-fujitsu: Original mock value for validationhandler was incorrect and not actually testing `EqaProviderName` logic, I updated mock of validationhandler with the values mentioned in my previous comment and added unit test to test `EqaProviderNam`e logic

ChrisJWoodcock (approved):

PR #1582, opened 2023-11-30
FAI-1012 - Add new banner for gov login migration

dashton82:


src/SFA.DAS.AssessorService.Web/Models/HomeIndexViewModel.cs
```diff
@@ -3,6 +3,6 @@
     public class HomeIndexViewModel
     {
         public bool UseGovSignIn { get; set; }
-        public string BannerViewPath => Constants.Banners.APARNotifyBannerViewPath;
+        public string BannerViewPath => Constants.Banners.GovNotifyBannerViewPath;
```
ChrisJWoodcock, line 6: If the APAR banner is going away - assuming that has been confirmed? - then get rid completly

ChrisJWoodcock (changes requested):

PR #1585, opened 2023-12-15
QF-1779 updating gettoken to gettokenasync

Jordan-fujitsu:


ChrisJWoodcock:
Code combined with existing PR for QF-687

PR #1587, opened 2024-01-05
QF-1083-Assessor-Internal-API-calls-QnA-API-using-MI-credentials

Jordan-fujitsu:


src/SFA.DAS.AssessorService.Application.Api.UnitTests/Controllers/ControllerAuthorizeAttributeTests.cs
```diff
@@ -15,7 +15,7 @@ public class ControllerAuthorizeAttributeTests
 
         private readonly List<string> _controllersThatDoNotRequireAuthorize = new List<string>()
         {
-            "PingController"
+            "PingController",
```
ChrisJWoodcock, line 18: An odd change - any value to this?

Jordan-fujitsu: Nope this was an accident, I wanted to test one of the endpoints on AT that uses the Qna api i temporarily made the endpoint (as i didnt have the client creds to hand to generate a bearer) to do this and added the controller to this test. When I reverted my changes back i accidently left in this comma, I hope that makes sense

ChrisJWoodcock: Sure

ChrisJWoodcock (dismissed):

PR #1588, opened 2024-01-08
QF-1794 Assessor code quality improvements

ChrisJWoodcock:


src/SFA.DAS.AssessorService.Api.Types/Models/MergeLogEntry.cs
```diff
@@ -5,11 +5,9 @@ namespace SFA.DAS.AssessorService.Api.Types.Models
     public class MergeLogEntry
     {
         public int Id { get; set; }
-        public string PrimaryEndPointAssessorOrganisationId { get; set; }
         public string PrimaryEndPointAssessorOrganisationName { get; set; }
-        public string SecondaryEndPointAssessorOrganisationId { get; set; }
         public string SecondaryEndPointAssessorOrganisationName { get; set; }
-        public DateTime SecondaryEPAOEffectiveTo { get; set; }
-        public DateTime? CompletedAt { get; set; }
+        public DateTime CompletedAt { get; set; }
```
Jordan-fujitsu, line 10: @ChrisJWoodcock is there a reason that `CompletedAt ` is no longer nullable?

pipeline-templates/job/deploy.yml
```diff
@@ -1,7 +1,7 @@
 parameters:  
   Environment:
   ServiceConnection:
-  LegacyApimServiceConnection: NAS Pre-Prod (ARM)
+  LegacyApimServiceConnection: NAS-PreProd-ARM (FCS)
```
Jordan-fujitsu, line 4: @ChrisJWoodcock why does this service connection needs to be updated?

src/SFA.DAS.AssessorService.Domain/Consts/CertificateActions.cs
```diff
@@ -8,8 +8,8 @@ public static class CertificateActions
         public const string ConfirmAddress = "ConfirmAddress";
         public const string Date = "Date";
         public const string SendTo = "SendTo";
-        public const string FirstName = "FirstName";
-        public const string LastName = "LastName";
+        public const string GivenNames = "GivenNames";
```
Jordan-fujitsu, line 11: @ChrisJWoodcock what is the reason for replacing the FirstName, LastName constants with GivenNames, FamilyName?

src/SFA.DAS.AssessorService.Infrastructure/ApiClients/QnA/IQnaApiClient.cs
```diff
@@ -7,17 +7,26 @@
...
 using System.Threading.Tasks;
 
-namespace SFA.DAS.AssessorService.Application.Api.Client.QnA
+namespace SFA.DAS.AssessorService.Infrastructure.ApiClients.QnA
 {
     public interface IQnaApiClient
     {
         Task<StartApplicationResponse> StartApplications(StartApplicationRequest startAppRequest);
-        Task<ApplicationData> GetApplicationData(Guid applicationId);
+        Task<T> GetApplicationData<T>(Guid applicationId);
```
Jordan-fujitsu, line 15: @ChrisJWoodcock whats the reason for making `GetApplicationData` a genric type ?

pipeline-templates/job/deploy.yml
ChrisJWoodcock: It's been done on master I'll sync the branches

src/SFA.DAS.AssessorService.Domain/Consts/CertificateActions.cs
ChrisJWoodcock: So that the admin service code matches the assessor code, the most correct code in this case was the version from the admin service.

src/SFA.DAS.AssessorService.Infrastructure/ApiClients/QnA/IQnaApiClient.cs
ChrisJWoodcock: I have synced the implementation of the QnAClient between admin and assessor in a follow up ticket the QnAClient which now has two separate but identical implementation will be migrated to the QnA API repo, and used in both place via a nuget, however when that method is called it serializes the dictionary to a known type which will be different from different callers - from admin and assessor it will be the same type when talking about assessor qna data but from RoATP it would be the RoATP definition of qna data

src/SFA.DAS.AssessorService.Api.Types/Models/MergeLogEntry.cs
ChrisJWoodcock: It never had a null value in it - and I needed the types to match between admin and assessor

Jordan-fujitsu (approved):

src/SFA.DAS.AssessorService.Api.Common/Settings/IAzureApiClientConfiguration.cs
```diff
@@ -1,11 +1,17 @@
...
+
         string Key { get; set; }
+
+        string ApiBaseAddress { get; set; }
+
         string ProductId { get; set; }
+
         string GroupId { get; set; }
+
+        string RequestBaseAddress { get; set; }
```
Jordan-fujitsu, line 15: @ChrisJWoodcock what is the difference between the` ApiBaseAddress` and `RequestBaseAddress` ?

pipeline-templates/job/code-build.yml
```diff
@@ -53,4 +53,9 @@ jobs:
     displayName: Publish Build Artifact
     inputs:
       targetPath: $(build.artifactstagingdirectory)/publish
-      artifactName: ${{ parameters.SolutionBaseName }}
\ No newline at end of file
+      artifactName: ${{ parameters.SolutionBaseName }}
+
+  - template: azure-pipelines-templates/build/step/nuget-pack.yml@das-platform-building-blocks
+    parameters:
+      DotNetStandardPackagesToPack: |
```
Jordan-fujitsu, line 60: @ChrisJWoodcock I assume this is the task to publish the shared nuget package to nuget

ChrisJWoodcock: Exactly

src/SFA.DAS.AssessorService.Api.Common/Settings/IAzureApiClientConfiguration.cs
ChrisJWoodcock: Just legacy with different develops calling things differently I tried to make as few changes as possible in this PR... there is still lots to do

Jordan-fujitsu (dismissed):

HussainsGitHub (approved):

PR #1596, opened 2024-01-22
QF-1079-Assessor-Internal-API-calls-RoATP-API-using-MI-credentials

Jordan-fujitsu:


src/SFA.DAS.AssessorService.Api.Common/Settings/RoatpApiClientConfiguration.cs
```diff
@@ -0,0 +1,9 @@
+﻿
+namespace SFA.DAS.AssessorService.Api.Common.Settings
+{
+    public class RoatpApiClientConfiguration : SFA.DAS.Http.Configuration.IManagedIdentityClientConfiguration
```
ChrisJWoodcock, line 4: This can be Settings as we are not going to share it with Assessor

src/SFA.DAS.AssessorService.Application.Api/StartupConfiguration/Startup.cs
```diff
@@ -244,15 +238,17 @@ private IServiceProvider ConfigureIOC(IServiceCollection services)
                 config.For<IReferenceDataTokenService>().Use<ReferenceDataTokenService>()
                     .Ctor<IClientConfiguration>().Is(Configuration.ReferenceDataApiAuthentication);
 
-                config.For<IRoatpTokenService>().Use<RoatpTokenService>()
-                    .Ctor<IClientConfiguration>().Is(Configuration.RoatpApiAuthentication);
+                config.For<SFA.DAS.Http.Configuration.IManagedIdentityClientConfiguration>().Use(Configuration.RoatpApiAuthentication);
```
ChrisJWoodcock, line 241: This won't work when we have more IManagedIdentityClientConfiguration derived classes it should config the RoatpApiClientConfiguration and then inject that into the RoatpApiClientFactory

src/SFA.DAS.AssessorService.Infrastructure/ApiClients/Roatp/IRoatpApiClientFactory.cs
```diff
@@ -0,0 +1,15 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Http;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace SFA.DAS.AssessorService.Infrastructure.ApiClients.Roatp
+{
+    public interface IRoatpApiClientFactory
```
ChrisJWoodcock, line 11: Is there a SFA.DAS.Http interface this can derive from because in the next set of ticket we will need to pass the factory into the base api client which a generic interface if following the factory pattern

ChrisJWoodcock (changes requested):

src/SFA.DAS.AssessorService.Application.Api/StartupConfiguration/Startup.cs
Jordan-fujitsu: Ah i see, good spot. Using RoatpApiClientConfiguration now and injecting that into the RoatpApiClientFactory

src/SFA.DAS.AssessorService.Infrastructure/ApiClients/Roatp/IRoatpApiClientFactory.cs
Jordan-fujitsu: There is `IHttpClientFactory` with the method  `HttpClient CreateHttpClient();` in SFA.DAS.Http which `IRoatpApiClientFactory` is now inheriting from

src/SFA.DAS.AssessorService.Infrastructure/ApiClients/Roatp/RoatpApiClientFactory.cs
```diff
@@ -0,0 +1,31 @@
...
+        public RoatpApiClientFactory(RoatpApiClientConfiguration roatpApiClientConfiguration) 
+        {
+            _roatpApiClientConfiguration = roatpApiClientConfiguration;
+        }
+
+
+        public HttpClient CreateHttpClient()
+        {
+            var httpClient = new ManagedIdentityHttpClientFactory(_roatpApiClientConfiguration).CreateHttpClient();
+            httpClient.DefaultRequestHeaders.Add("Accept", "Application/json");
```
ChrisJWoodcock, line 26: The header is already added by the SFA.DAS.Http.ManagedIdentityHttpClientFactory via its HttpBuilder and DefaultHeadersHandler

ChrisJWoodcock (changes requested):

src/SFA.DAS.AssessorService.Infrastructure/ApiClients/Roatp/RoatpApiClientFactory.cs
Jordan-fujitsu: Removed code

ChrisJWoodcock (approved):

PR #1597, opened 2024-01-29
QF-1752 Correct get version and standard version from lars code and add integration tests

ChrisJWoodcock:


src/SFA.DAS.AssessorService.Database/Functions/Scalar-valued Functions/GetStandardUidFromLarsCode.sql
alandburns, line 1: The simpler fix is to change the 
ORDER BY VersionMajor DESC, VersionMinor DESC 
to be
ORDER BY VersionMajor, VersionMinor

src/SFA.DAS.AssessorService.Database/Functions/Scalar-valued Functions/GetVersionFromLarsCode.sql
alandburns, line 1: The simpler fix is to change the 
ORDER BY VersionMajor DESC, VersionMinor DESC 
to be
ORDER BY VersionMajor, VersionMinor

Jordan-fujitsu (approved):

Jordan-fujitsu (approved):

PR #1598, opened 2024-02-07
QF-1829 Assessor MI changes

ChrisJWoodcock:


src/SFA.DAS.AssessorService.Web/Startup.cs
```diff
@@ -229,7 +200,7 @@ private IServiceProvider ConfigureIoc(IServiceCollection services)
             {
                 config.Scan(_ =>
                 {
-                    _.AssemblyContainingType(typeof(Startup));
+                    _.AssembliesFromApplicationBaseDirectory(c => c.FullName.StartsWith("SFA"));
```
Jordan-fujitsu, line 203: @ChrisJWoodcock is this registration any services which the namespace starts with SFA?

src/SFA.DAS.AssessorService.Web/Helpers/EpaOrgIdFinder.cs
```diff
@@ -8,6 +8,6 @@ public static class EpaOrgIdFinder
 
     public static string GetFromClaim(IHttpContextAccessor accessor)
     {
-        return accessor.HttpContext.User.FindFirst(ClaimType)?.Value;
+        return accessor?.HttpContext?.User?.FindFirst(ClaimType)?.Value;
```
Jordan-fujitsu, line 11: @ChrisJWoodcock is there a reason why this has been updated?

src/SFA.DAS.AssessorService.Web/appsettings.json
```diff
@@ -17,5 +17,6 @@
   "ConnectionStrings": {
     "Redis": ""
   },
-  "EnvironmentName": "LOCAL"
+  "EnvironmentName": "LOCAL",
+  "StubAuth": true
```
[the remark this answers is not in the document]
ChrisJWoodcock: The StubAuth when enabled is a local workaround for the GOV.UK One Login, when enabled it prompt you to enter details - a SignInId and Email which are in the Contacts table and then you are that user when you logon

src/SFA.DAS.AssessorService.Web/Startup.cs
ChrisJWoodcock: Yes - this is the same as used in the Api and External API already

Jordan-fujitsu (changes requested):
@ChrisJWoodcock having issues running locally, seems like a DI error can you double check?

ChrisJWoodcock:
> @ChrisJWoodcock having issues running locally, seems like a DI error can you double check?

This is not a comment for a PR

Jordan-fujitsu (approved):

PR #1602, opened 2024-03-05
QF-1829 Add app role assignments to external api and external sandbox api

ChrisJWoodcock:


pipeline-templates/job/deploy.yml
```diff
@@ -90,6 +90,12 @@ jobs:
             ResourceName: $(apiAppServiceName)
             Tenant: $(Tenant)
             IsMultiRepoCheckout: true
+        - template: azure-pipelines-templates/deploy/step/app-role-assignments.yml@das-platform-building-blocks
+          parameters:
+            ServiceConnection: ${{ parameters.AppRoleAssignmentsServiceConnection }}
+            ResourceName: $(externalApiAppServiceName)
```
debrr07, line 96: ResourceName needs changing to ExternalApiAppServiceName

ChrisJWoodcock: @debrr07 Ok corrected that and the ApiAppServiceName also for consistency but the "apiAppServiceName" having been incorrect for some time suggests these are not actually case sensitive

debrr07 (approved):

PR #1603, opened 2024-03-18
Qf 1839 remove login service legacy code

LACrossley:
Removed defunct legacy code pertaining to the switch to GovUkSignIn. All code now defaults to use Gov UK Sign In and the switch enabling the DfE internal LoginService has been removed.


src/SFA.DAS.AssessorService.Application/Handlers/ContactHandlers/CreateContactHandler.cs
```diff
@@ -80,25 +80,9 @@ public async Task<ContactBoolResponse> Handle(CreateContactRequest createContact
...
             else
             {
-                if (!string.IsNullOrEmpty(createContactRequest.GovIdentifier))
-                {
-                    await _contactRepository.UpdateSignInId(existingContact.Id, Guid.NewGuid(), createContactRequest.GovIdentifier);
-                    response.Result = true;
-                    return response;
-                }
-                
-                var invitationResult = await _signInService.InviteUser(createContactRequest.Email, createContactRequest.GivenName, createContactRequest.FamilyName, existingContact.Id);
```
ChrisJWoodcock, line 90: Why is this section removed?

src/SFA.DAS.AssessorService.Web/Controllers/AccountController.cs
```diff
@@ -232,21 +232,13 @@ public IActionResult UpdateAnAccount()
...
-            }
-            var vm = new CreateAccountViewModel();
-            return View(vm);
+            return RedirectToAction("UpdateAnAccount");
         }
 
         [HttpPost]
         public async Task<IActionResult> CreateAnAccount(CreateAccountViewModel vm)
-        {
-            if (_config.UseGovSignIn)
```
ChrisJWoodcock, line 246: The code after the redirect will never be executed, and posting back now always redirects to an error - how does the post get called - it would never be a good idea to post back now - can it be posted back - if so then need to work out what to do with that code

src/SFA.DAS.AssessorService.Web/Startup.cs
```diff
@@ -85,18 +85,11 @@ public IServiceProvider ConfigureServices(IServiceCollection services)
...
-                {
-                    var isLocal = string.IsNullOrEmpty(_config["ResourceEnvironmentName"])
-                                  || _config["ResourceEnvironmentName"].Equals("LOCAL", StringComparison.CurrentCultureIgnoreCase);
-                    var cookieDomain = isLocal ? "" : Configuration.ServiceLink.Replace("https://", "", StringComparison.CurrentCultureIgnoreCase);
-                    var loginRedirect = isLocal ? "" : $"{Configuration.ServiceLink}/service/account-details";
-                    services.AddAndConfigureGovUkAuthentication(_config, typeof(AssessorServiceAccountPostAuthenticationClaimsHandler), "/account/signedout", "/service/account-details", cookieDomain, loginRedirect);
-                }
-                else
-                {
-                    services.AddAndConfigureAuthentication(Configuration, _env);
```
ChrisJWoodcock, line 98: I don't see AddAndConfigureAuthentication being removed - wouldn't it now be redundant?

src/SFA.DAS.AssessorService.Web/StartupConfiguration/AuthenticationStartup.cs
```diff
@@ -52,9 +52,7 @@ public static class AuthenticationStartup
                     };
 
                     options.SignInScheme = "Cookies";
-                    options.Authority = _configuration.LoginService.MetadataAddress;
```
ChrisJWoodcock, line 55: Yes - here is the method with a couple of lines removed - but what else calls this?

src/SFA.DAS.AssessorService.Web/Views/Banners/_GovNotifyBannerPartial.cshtml
```diff
@@ -1,7 +1,7 @@
 ﻿<div class="govuk-notification-banner" role="region"
      aria-labelledby="govuk-notification-banner-title"
      data-module="govuk-notification-banner">
-    <div class="govuk-notification-banner__header">
```
ChrisJWoodcock, line 4: With a content change like this need to confirm with PO and Content designer that the message is being removed

src/SFA.DAS.AssessorService.Web/Views/Home/Index.cshtml
```diff
@@ -4,7 +4,7 @@
     ViewData["Title"] = "Apprenticeship assessment service";
     ViewBag.ZenDeskLabel = "ass-start-page";
     Layout = "_Layout";
-    var actionName = Model.UseGovSignIn ? "UpdateAnAccount" : "CreateAnAccount";
```
ChrisJWoodcock, line 7: This a link to CreateAnAccount being removed - could that be the only call to CreateAnAccount if so see the comment above about the Post

ChrisJWoodcock (changes requested):

src/SFA.DAS.AssessorService.Web/Controllers/AccountController.cs
LACrossley: If the Post CreateAnAccount method simply redirects to an error page and the CreateAnAccount view calls the Post method, then neither the Post method nor the view are required anymore. Would you agree?

I think the Get CreateAnAccount method is called by the "back" button on the InviteSent view. Since that always now redirects to the Get UpdateAnAccount method, surely it would make sense to replace the view's call with a direct one and delete the Get Create method from the controller?

Are there any cases where we still manage identities internally? I can see there are some code paths from the InviteSent process which don't obviously depend on the GovSignIn variable. The InviteSent action calls upon the CreateAccountViewModel. What is the purpose of the InviteSent process? Would it make sense to rename the view model to "Update..." rather than "Create..." or do you think we may be able to delete the InviteSent code paths?

I can't see anywhere where the Get InviteSent method is called. Is it called by any external processes, do you think?

src/SFA.DAS.AssessorService.Web/Startup.cs
LACrossley: I think I've removed the else clause that calls AddAndConfigureAuthentication, and left in the ...GovUk... equivalent. Do you mean I should remove the method definition from the AuthenticationStartup class?

Wouldn't that mean the whole AS class is now redundant?

src/SFA.DAS.AssessorService.Web/StartupConfiguration/AuthenticationStartup.cs
LACrossley: Nothing that I can see. So the whole class is redundant now.

src/SFA.DAS.AssessorService.Web/Models/HomeIndexViewModel.cs
```diff
@@ -2,7 +2,6 @@
 {
     public class HomeIndexViewModel
     {
-        public bool UseGovSignIn { get; set; }
```
ChrisJWoodcock, line 5: Compile errors here

ChrisJWoodcock (changes requested):

ChrisJWoodcock (approved):

PR #1616, opened 2024-05-17
QF-1912-Update_pending_users

karen-hanna:
added script to update Pending users to Live


src/SFA.DAS.AssessorService.Database/PostDeploymentScripts/QF-1912-AddIndexToContacts-GovUkIdentifier.sql.sql
```diff
@@ -0,0 +1,4 @@
+﻿/*
+	Add an index to Contacts.GovUkIdentifier column
+*/
+CREATE INDEX [IX_Contacts_GovUkIdentifier] ON [Contacts] (GovUkIdentifier)
```
ChrisJWoodcock, line 4: This should be added to the [Contacts] table definition not a post deployment script - this will try and create an index each time it is run - first time would be okay, second time will fail - by adding to the [Contacts] table definition the DACPAC will sync it correctly.

src/SFA.DAS.AssessorService.Database/PostDeploymentScripts/QF-1912-UpdatePendingUsersToLive.sql
```diff
@@ -0,0 +1,12 @@
+﻿/*
+	Update User Status to Live for users with Pending Status and GovUkIdentifier is not null or empty
```
ChrisJWoodcock, line 2: Explain that this is a temporary script which when deployed to all environments after QF-1784 will correct the existing data and then this script can be removed.

ChrisJWoodcock (changes requested):

src/SFA.DAS.AssessorService.Database/PostDeploymentScripts/QF-1912-AddIndexToContacts-GovUkIdentifier.sql.sql
karen-hanna: Moved index to table definition

src/SFA.DAS.AssessorService.Database/PostDeploymentScripts/QF-1912-UpdatePendingUsersToLive.sql
karen-hanna: Updated description in update script file and post deployment script file

ChrisJWoodcock (approved):

ChrisJWoodcock (approved):

PR #1645, opened 2025-03-12
P2-2402 Added get assessments endpoint

ChrisJWoodcock:


src/SFA.DAS.AssessorService.Domain/SFA.DAS.AssessorService.Domain.csproj
```diff
@@ -5,12 +5,14 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="FluentValidation.AspNetCore" Version="11.3.0" />
-    <PackageReference Include="newtonsoft.json" Version="13.0.3" />
+    <Compile Remove="DTOs\Register\**" />
+    <EmbeddedResource Remove="DTOs\Register\**" />
+    <None Remove="DTOs\Register\**" />
```
karen-hanna, line 10: Do we need these lines?  I can't see a  Register folder in this location?

karen-hanna (approved):

karen-hanna (approved):

PR #1647, opened 2025-03-20
P2 2363 Framewrok learner details after AS reprint

karen-hanna:


src/SFA.DAS.AssessorService.Data/Staff/StaffCertificateRepository.cs
```diff
@@ -135,6 +135,34 @@ ORDER BY
             return result;
         }
 
+        public async Task<List<CertificateLogSummary>> GetLatestCertificateLogs(Guid certificateId, int count)
```
ChrisJWoodcock, line 138: This is the same as GetLatestCertificateLogs - just with a count - so you should add a count to the existing method and default it to 1

ChrisJWoodcock (changes requested):

ChrisJWoodcock (approved):

PR #1669, opened 2025-10-24
P2-2948 - Increased varchar column to 32

philhead1:


src/SFA.DAS.AssessorService.Database/Tables/OrganisationStandard.sql
alandburns, line 1: The column in the main tables should remain unchanged, only the Staging table needs to be amended and you need to ensure that only valid IfateReferenceNumber / StandardReference value is stored to the OrganisationStandard table.
the transfer from Staging should work anyway as it has to join to Standards table, which only contains valid values.

ChrisJWoodcock:
This has been superseded by P2-2946

PR #1670, opened 2025-11-14
CSP-2315 Refactored RoatpClient to use new endpoints

DevenKShah:


src/SFA.DAS.AssessorService.Infrastructure/ApiClients/Roatp/Types/Organisation.cs
```diff
@@ -1,19 +1,19 @@
...
 {
     using System;
 
     public class Organisation
     {
         public Guid Id { get; set; }
         public ProviderType ProviderType { get; set; }
-        public OrganisationType OrganisationType { get; set; }
+        public int OrganisationTypeId { get; set; }
+        public string OrganisationType { get; set; }
```
ChrisJWoodcock, line 10: Why does OrganisationType use a string but the ProviderType use an enum?

src/SFA.DAS.AssessorService.Database/StoredProcedures/DatabaseMaintenance.sql
```diff
@@ -4,47 +4,47 @@ BEGIN
 	DECLARE @Start DATETIME = GETUTCDATE()
 	DECLARE @Results TABLE(Step VARCHAR(100))
 
-	ALTER INDEX ALL ON [dbo].[Certificates]
-	REBUILD WITH
-	(
-	    FILLFACTOR = 80,
-	    ONLINE = ON
-	);
+	--ALTER INDEX ALL ON [dbo].[Certificates]
```
ChrisJWoodcock: Why are you commenting these out? I think the database maintenance is required - DevOps said not when I asked them 4 years ago, they claims no Azure database needed specific index rebuilds because it does the job itself, however experimentation showed that rebuilding these indexes periodically results in a dramatic increase in performance

ChrisJWoodcock (changes requested):

src/SFA.DAS.AssessorService.Database/StoredProcedures/DatabaseMaintenance.sql
DevenKShah: That was me trying to get it working locally, I didn't mean to commit this. I will roll it back.

src/SFA.DAS.AssessorService.Infrastructure/ApiClients/Roatp/Types/Organisation.cs
DevenKShah: ProviderType is fixed and tiny, where as Org type is long and varies based on provider type.

ChrisJWoodcock (dismissed):

ChrisJWoodcock (approved):

PR #1673, opened 2025-12-02
P2 2946 ofqual data upload

saqlainmalik2000:


src/SFA.DAS.AssessorService.Data.IntegrationTests/SFA.DAS.AssessorService.Data.IntegrationTests.csproj
```diff
@@ -34,6 +34,9 @@
   </ItemGroup>
 
   <ItemGroup>
+    <None Update="connectionStrings.Local.json">
```
ChrisJWoodcock: This is duplicated

ChrisJWoodcock (changes requested):

src/SFA.DAS.AssessorService.Data.IntegrationTests/SFA.DAS.AssessorService.Data.IntegrationTests.csproj
saqlainmalik2000: Removed duplication

ChrisJWoodcock (approved):

PR #1677, opened 2026-03-12
P2-2558-Request print

ratheeshri1984:


src/SFA.DAS.AssessorService.Application/Handlers/Certificates/UpdateCertificatePrintRequestHandler.cs
```diff
@@ -0,0 +1,63 @@
...
+            }
+
+            if (!string.Equals(certificate.LatestEPAOutcome, EpaOutcome.Pass, StringComparison.InvariantCultureIgnoreCase) ||
+                !string.Equals(certificate.Status, CertificateStatus.Submitted, StringComparison.InvariantCultureIgnoreCase) ||
+                certificate.PrintRequestedAt != null)
+            {
+                throw new ArgumentException("Certificate is not in a valid state to request print");
+            }
+
+            certificate.CertificateData.ContactName = request.Address.ContactName;
```
ChrisJWoodcock, line 42: The cert date is not checked for null

src/SFA.DAS.AssessorService.Application/Handlers/Certificates/UpdateCertificatePrintRequestHandler.cs
```diff
@@ -0,0 +1,63 @@
...
+            certificate.CertificateData.ContactAddLine2 = request.Address.ContactAddLine2;
+            certificate.CertificateData.ContactAddLine3 = request.Address.ContactAddLine3;
+            certificate.CertificateData.ContactAddLine4 = request.Address.ContactAddLine4;
+            certificate.CertificateData.ContactPostCode = request.Address.ContactPostCode;
+
+            certificate.PrintRequestedAt = request.PrintRequestedAt;
+            certificate.PrintRequestedBy = request.PrintRequestedBy;
+            certificate.Status = CertificateStatus.PrintRequested;
+
+            var actor = certificate.CertificateData?.ContactName ?? request.Address?.ContactName ?? string.Empty;
```
ChrisJWoodcock: The cert date is checked for null and then if not gets the vale which was just put there

ChrisJWoodcock (changes requested):

src/SFA.DAS.AssessorService.Application/Handlers/Certificates/UpdateCertificatePrintRequestHandler.cs
ratheeshri1984: "latestEPAOutcome": "Pass"
"status": "Submitted"
"printRequestedAt": null    This is the criteria mentioned in the technical document. Do we need to include AchievementDate?

src/SFA.DAS.AssessorService.Application/Handlers/Certificates/UpdateCertificatePrintRequestHandler.cs
ratheeshri1984: Done

src/SFA.DAS.AssessorService.Application/Handlers/Certificates/UpdateCertificatePrintRequestHandler.cs
ChrisJWoodcock: No I was commenting on CertifcateData null checks before setting it

ChrisJWoodcock (approved):

PR #1681, opened 2026-03-25
P2-2559 -Epa_Filter

ratheeshri1984:


src/SFA.DAS.AssessorService.Data/CertificateRepository.cs
```diff
@@ -166,7 +166,7 @@ public async Task<List<ApprenticeCertificateSummary>> GetPrintableCertificates(l
             var statusesToExclude = new[] { CertificateStatus.Draft, CertificateStatus.Deleted };
 
             var standardCertificates = await _unitOfWork.AssessorDbContext.StandardCertificates
-                .Where(c => c.Uln == uln && !statusesToExclude.Contains(c.Status) && c.AchievementDate != null)
+                .Where(c => c.Uln == uln && !statusesToExclude.Contains(c.Status) && c.AchievementDate != null && c.LatestEPAOutcome == EpaOutcome.Pass)
```
ChrisJWoodcock, line 169: Need additional test on this to check that some are now excluded:

certificate with AchievementDate set and LatestEPAOutcome = null is excluded
certificate with LatestEPAOutcome = Fail is excluded
certificate with LatestEPAOutcome = Pass is included

ChrisJWoodcock (changes requested):

src/SFA.DAS.AssessorService.Data/CertificateRepository.cs
ratheeshri1984: Done

ChrisJWoodcock (approved):

PR #1693, opened 2026-05-20
P2 3244 dashboard performance improvements

alandburns:
Rework SP GetEPAO_DashboardCounts to improve performance


src/SFA.DAS.AssessorService.Database/StoredProcedures/GetEPAO_DashboardCounts.sql
```diff
@@ -3,22 +3,55 @@ CREATE PROCEDURE [GetEPAO_DashboardCounts]
...
+	AND EndPointAssessorOrganisationId = @epaOrgId
+	;
+
 	SELECT SUM(Standards) Standards, SUM(Pipeline) Pipeline, SUM(Assessments) Assessments
 	FROM (
 		-- The active records from ilr
 		SELECT 0 Assessments, COUNT(*) Pipeline, 0 Standards
-		FROM [dbo].[EPAO_Func_Get_PipelineInfo] (@epaOrgId, NULL, @pipelineCutOff)
+		FROM [dbo].[Learner] le1
+		JOIN #LiveStandards lv1 on lv1.EndPointAssessorOrganisationId = le1.EpaOrgId AND lv1.StandardCode = le1.StdCode
```
ChrisJWoodcock, line 24: The function is reused multiple times, its a TVF so should be expanded inline, removing it is a maintenance issue as the logic can now diverge - seems more likely that the performance issue is caused by a missing index e.g.

```
CREATE NONCLUSTERED INDEX IX_Learner_PipelineInfo
ON dbo.Learner (EpaOrgId, CompletionStatus, StdCode, EstimatedEndDate, LastUpdated)
INCLUDE (Uln, UkPrn, StandardReference, StandardName, Version);
```
Where the EpaOrgId is the first discriminator rather than the last as in the existing index - did you rule out a missing index?

ChrisJWoodcock (changes requested):

src/SFA.DAS.AssessorService.Database/StoredProcedures/GetEPAO_DashboardCounts.sql
alandburns: This SP is just for the Dashboard, and it is doing too much work by using the function. And I am not bothered about the maintenance issue as this is trivial code.

PR #1697, opened 2026-06-03
P2-3108-dotnet-upgrade-10

srinivasdudam-dfe:


src/SFA.DAS.AssessorService.Web/StartupConfiguration/Authorization/AssessorServiceAccountPostAuthenticationClaimsHandler.cs
```diff
@@ -65,14 +71,6 @@ public async Task<IEnumerable<Claim>> GetClaims(TokenValidatedContext tokenValid
                 {
                     _logger.LogInformation("Failed to retrieve user by email.");
                 }
-
```
ChrisJWoodcock, line 68: Need to explain this

src/SFA.DAS.AssessorService.Api.Types/Models/Standards/StandardVersion.cs
```diff
@@ -24,6 +24,22 @@ public class StandardVersion
         public bool EPAChanged { get; set; }
         public string StandardPageUrl { get; set; }
         public bool EpaoMustBeApprovedByRegulatorBody { get; set; }
+        public int VersionMajor { get; set; }
```
ChrisJWoodcock: This is emitting more information from the API - is there an alternative here?

src/SFA.DAS.AssessorService.Api.Common/OpenApiStubs.cs
```diff
@@ -0,0 +1,46 @@
+// Minimal OpenAPI model stubs to restore build after package upgrades.
+// These are lightweight placeholders mirroring only the members used by the codebase.
+namespace Microsoft.OpenApi.Models
```
ChrisJWoodcock: This is odd - need you to explain it

ChrisJWoodcock (changes requested):

src/SFA.DAS.AssessorService.Api.Types/Models/Standards/StandardVersion.cs
srinivasdudam-dfe: Added the properties in StandardVersion.cs to fix a Unit test
SearchStandardsBySearchTermTests --->ResultsMatchExpectedListOfStandars test which is trying to compare a domain Standard object and Model Standard object. In Domain Standard object which hase these properties VersionMajor, VersionMinor etc.

src/SFA.DAS.AssessorService.Api.Common/OpenApiStubs.cs
srinivasdudam-dfe: in test/at i have to check if the Bearer token is working from swagger

src/SFA.DAS.AssessorService.Api.Types/Models/Standards/StandardVersion.cs
srinivasdudam-dfe: revert the code and see if the tests pass

src/SFA.DAS.AssessorService.Web/StartupConfiguration/Authorization/AssessorServiceAccountPostAuthenticationClaimsHandler.cs
srinivasdudam-dfe: we decided to rearrange the code but the functionality remains the same.
Added a new claim ContactStatus when the user.Status is deleted.  And check this clain in the calling method and redirect to Accesdenied page

src/SFA.DAS.AssessorService.Api.Types/Models/Standards/StandardVersion.cs
srinivasdudam-dfe: reverted code and all unit test pass

